PR #17, opened 2015-03-29
Windows-related fixes.

rdoeffinger:
The patches I sent as issues before as pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ TODO
 1. ~~non-blocking actions~~
 2. ~~multi-lingual~~
 3. ~~filtering and autocomplete~~
-4. edit, insert
 5. gpg-id management (per-folder)
 
 Instalation
--- a/dialog.cpp
+++ b/dialog.cpp
@@ -17,7 +17,6 @@ Dialog::Dialog(QWidget *parent) :
  */
 Dialog::~Dialog()
 {
-    delete ui;
 }
 
 /**
--- a/dialog.h
+++ b/dialog.h
@@ -49,7 +49,7 @@ private slots:
     void on_checkBoxAutoclear_clicked();
 
 private:
-    Ui::Dialog *ui;
+    QScopedPointer<Ui::Dialog> ui;
     void setGroupBoxState();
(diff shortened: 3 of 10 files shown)


rdoeffinger:
Sorry, was too lazy to create branches. This pull request now also contains edit/insert/delete feature implementation.
If you prefer I can make a branch with only the Windows fixes.

annejan:
Adding now, will replace popups with multiline edit (for password) and merge ..

rdoeffinger:
Thanks!
Just to make sure you didn't miss it: Since bbbdbe4 the password input dialog is multi-line if you use Qt 5.2 or later.
So there should be no need to replace it unless you want to fully support earlier versions (I kept the single line edit as fallback since I wasn't sure dropping support for such recent versions was reasonable).

annejan:
Ah yes, noticed it later on my main laptop (with up-to-date Qt).
In the develop branch I have made some changes last month which I would like to intergrate that used the multi-line field used for viewing, for editing too.
So I might still change the editing to a smoother (no popup) experience ;)
Thanks for your work, this has gotten me back into working on this project some more too :)

rdoeffinger:
Just for reference, I had (besides lazyness of course ;) ) a few half-good reasons not to use the existing field.
For example I felt that it is fairly overloaded already, since both errors, error messages and decrypted output end up in there. This seems particularly bad when the editing/adding/encrypting failed and you need to display the error in the same thing where you just entered the text...
Then specifically for editing, there is a bit of complex interaction with the clipboard features that remove e.g. the password. Would you when editing ignore that setting and display the full text?
Also some people might want an option to never show passwords, with the dialog it is fairly simple to just pop it up twice and ask to enter the same thing.
I'm sure it's quite possible to come up with something better than I implemented, it just felt like it might have to come with a bigger redesign of the interface.
Btw: one small thing I noticed: "pass" seems to add an extra newline at the end of the passwords, and it clearly assumes it to exist when it prints the value. I don't really like that behaviour, but it might make sense to align qtpass with that anyway....

PR #19, opened 2015-04-04
Support encrypting to multiple recipients.

rdoeffinger:
"pass" already supports that, and uses the one-line-per-recipient
format for .gpg-id.


annejan:
Thanks for fixing this

PR #22, opened 2015-04-04
More proper support for subdirectories.

rdoeffinger:
Creates new file in selected sub-directory.
Also searches for .gpg-id in current sub-directory
first.

Signed-off-by: Reimar Döffinger <Reimar.Doeffinger@gmx.de>


annejan:
Thanks, seems like a nice improvement

PR #23, opened 2015-04-04
Support for editing .gpg-id via GUI with public keyring list.

rdoeffinger:
This is not really a pull request since it does nothing useful.
But i don't know a better idea to make sure people are aware, and give a chance to comment early on.
I've been working a bit on making a user-interface to edit the .gpg-id file.
What I managed so far is creating a dialog with a list of all public keys and checkboxes for them.
The next step would be to read in an existing .gpg-id, set checkboxes according to that, add extra entries if a public key for a user is not available and after the dialog exits writing a new .gpg-id file.
There is a bit of a user-interface issue though, as the current UI makes it look as it is editable per-file, plus I'd prefer to not add code to change all existing files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ TODO
 2. ~~multi-lingual~~
 3. ~~filtering and autocomplete~~
 4. ~~edit, insert~~
-5. gpg-id management (per-folder)
+5. ~~gpg-id management (per-folder)~~
 
 Instalation
 -----------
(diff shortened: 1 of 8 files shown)


rdoeffinger:
While some of the issues I pointed out still exist (and yes, the usability really isn't any good), the code now works and I think it works well enough.
One known issue: if there are keys in the .gpg-id that aren't in the public key-ring I suspect bad things might happen. I have not yet tested it.
There are likely to be a few more issues like that, though I think it is already useful.

PR #46, opened 2015-05-09
Enable C++11 and use it to simplify loops.

rdoeffinger:
Not sure if it's worth it/a good idea, e.g. if there are compatibility concerns.
In particular not tested with Qt4.
Still, C++11 has quite a lot of nice features...


annejan:
Works well on my main (virtual) machines, will test this on "old" Qt4 setup tonight.

annejan:
This gave no troubles on any of my serious testing setups.
The only system it would break on I could find was CentOS 6.6 (gcc 4.4)

PR #78, opened 2015-08-02
[WIP] Call 'pass git init' on creation of password-store when useGit

dennisdegreef:
Whenever the `~/.password-store` does not exist, the wizard is ran to initialize it.
Depending on either already existing configuration or default configuration, `useGit` can be true.
If so, initialize the directory with git as well, else the `git commit` of the `.gpg-id` file will fail.

TODO:
- [ ] Remove qDebug statements in squash
- [ ] Make `executePassGitInit` configurable with either `pass` or `git+gpg`
- [ ] Seem to have problems with a 'command already running?'

PS. I'm not familiar with either QtPass' codebase, the language, or Qt in general, so all feedback is very welcome.

### DIFF
--- a/dialog.cpp
+++ b/dialog.cpp
@@ -495,7 +495,7 @@ void Dialog::wizard()
     }
 
     QStringList names = mainWindow->getSecretKeys();
-    //qDebug() << names;
+    qDebug() << names;
     if (QFile(gpg).exists() && names.empty()) {
         KeygenDialog d(this);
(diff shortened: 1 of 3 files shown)


dennisdegreef:
Erm... This was work in progress? For me, it still threw errors about multiple commands running?

annejan:
Monday morning merges . . 

.. it worked for me :)

PR #87, opened 2015-08-06
Issue 86 clear panel

karlgrz:
Implemented a "Autoclear panel" checkbox and corresponding duration control in the Config tab.

qtpass_config

Then, when the duration has passed, the content panel is cleared, like so:

qtpass_cleared

I also updated .gitignore for the artifacts generated from building on ubuntu (I was just running `qmake && make` in the qtpass root and then running the binary generated like `./qtpass` if there's a better, cleaner way let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,4 @@ qtpass.xcodeproj/xcuserdata/*
 qtpass.xcworkspace/xcuserdata/*
 .DS_Store
 .qmake.stash
-
+*.o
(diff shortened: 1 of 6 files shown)


annejan:
Wow, thanks, that's fast :)

Will give it a quick one-over and add Dutch and German translations for it 

I was never really bothererd by the artifacts (on OSX or Arch), but it's also a good addition to the .gitignore

I don't think there is really a cleaner way to do things.

annejan:
Looks great on first viewing and reading the diffs (on mobile) other thing I'll probably add are a sane default value for first time users and people upgrading.. 
Will merge this later today.

karlgrz:
Double bonus!

PR #103, opened 2015-09-28
spanish translations added

mrpnkt:
translated some strings


annejan:
Thanks for the effort!

PR #104, opened 2015-10-06
Get PREFIX variable from environment

jorti:
The parentheses are needed to get the variable from the environment


annejan:
This "fix" broke installation on Debian and CentOS (not tested on others)

annejan:
Will look into getting the environment variable in separate issue https://github.com/IJHack/qtpass/issues/106

The install currently only adheres to the Qt "standard" of `qtpass PREFIX=/path/to/my/bin`

PR #113, opened 2015-11-10
Fix bug that prints "Unknown error" to the terminal

dvarum12:
Then I try to start qtpass I get the message in the terminal "Unknown error" and the program closes, but this patch fixes that problem.
A workaround for this bug is to add a argument to the execution of qtpass, like "qtpass test". You can still use arguments to search in qtpass then starting the program.


annejan:
Awesome, will merge this later today

PR #119, opened 2015-11-21
Modified the clipboard logic to allow for on-demand copy to clipboard.

jonhanks:
Sometimes I want secondary information from a entry (URL, notes, ...) and don't always want the password being pushed to the clipboard.  So I added a new state to the clipboard handling, copy on-demand.

For the UI I changed the clipboard config entry to be a combobox instead of a checkbox.  It has 3 states:
 * Never copy to the clipboard
 * Always copy to the clipboard
 * On-demand copy to the clipboard

Added a 'Password' button to the main window which copies the password to the clipboard when pressed.
The button is disabled when there is no password loaded or passwords are set to never copy to the clipboard.

Added a context menu entry to the text browser to copy the password to the clipboard.

Moved some of the clipboard code around to make it more flexible.  Clippedpass is always set when the password could be copied.

### DIFF
--- a/configdialog.cpp
+++ b/configdialog.cpp
@@ -22,6 +22,12 @@ ConfigDialog::ConfigDialog(MainWindow *parent) :
     ui->profileTable->verticalHeader()->hide();
     ui->profileTable->horizontalHeader()->setStretchLastSection(true);
     ui->label->setText(ui->label->text() + VERSION);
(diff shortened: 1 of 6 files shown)


annejan:
Thanks for the effort.
This looks and feels like a very useful addition.

PR #120, opened 2015-11-23
Cleanup and coding standards

annejan:
Just some linting


annejan:
As clean as it's gonna get for now . . 
Mostly things I don't like ( "./something.h" instead of just "something.h" ) and Qt specific contradictions.

PR #129, opened 2015-12-31
qrand always generating the same sequence of passwords

treat1:
qrand fallback when not using pwgen would always generate the same
sequence of passwords because qsrand was never called.


annejan:
Great find!
Thanks for this fix...

PR #124, opened 2015-12-17
System Icons on Buttons and Doubleclick on treeView

jounathaen:
The Buttons "Add", "User", ... are now drawn with the system theme default icons and I added some spacing. (looks good on cinnamon 2.8)
Also  I added the functionality that a double-click on a item opens the edit window.
Best regards
Jonathan

PR #145, opened 2016-01-22
Pre 1.1 mixing

annejan:
This is mostly a mergein of #124 by @jounathaen and assets by @jsimplicio

Also trying out if this is a good strategy of opening a pull-request early.


annejan:
screen shot 2016-01-22 at 16 18 26
Some fallback icons as PoC
screen shot 2016-01-22 at 17 51 58
Almost as sexy as the native ones :smile:

annejan:
KDE screenshot_20160122_004108
Gnome 
screen shot 2016-01-22 at 18 09 39

jsimplicio:
Here are the icons without that button background. I think they'll look best now without that background in actually all interfaces. Let me know what you think. I also changed the update icon because an arrow down doesn't really make much sense for update? 

[qtpass-icons-nobg.zip](https://github.com/IJHack/qtpass/files/101220/qtpass-icons-nobg.zip)

annejan:
Technically it's `git push` and `git pull` the update is still from the first (read only) version.

Thanks, will have some time to play with then after diner I hope :smile:

jsimplicio:
@annejan Oh man so it's pull and push I had no idea. I see. Let me fix it and turn it back to a down arrow then, it makes more sense.

Here is the final set with that pull fixed.

[qtpass-icons-final.zip](https://github.com/IJHack/qtpass/files/101238/qtpass-icons-final.zip)

jsimplicio:
Just want to recommend using https://github.com/svg/svgo in the future for the icons I've made. If you read this tools description it removes useless and redundant excess info that is exported on the svg when you export it from a program. I used Sketch and it added a lot of fluff to the files.

annejan:
Done :grin:

jsimplicio:
:+1:

jounathaen:
Thanks for merging! Also nice Icons!

annejan:
Thank you for the awesome effort!
And finding out how the (standard) theme icons work.

I really love how it looks on KDE Plasma.
On Gnome it looks gnomey and fun.
Still wan

And the fallback icons work well on Win and Mac indeed :smile:

jounathaen:
I did a clean pull, but for some reason, the Icons don't work for me. Instead the folder Icons in the treeview have changed to some non system Theme (I use cinnamon). I'll check if this is a problem of my setup...
Now:

Before:
screenshot from 2016-01-22 23 30 07

(Maybe this is all a problem of my Linux Installation)
BTW: When I do a make uninstall, it tries to remove my /usr/bin/ which I really want to keep :-D

annejan:
Wow!! 

I'll have a quick look on some more systems when I get home.

It should show your system icons where possible.
There might be some "wrong" still. 

Should currently:
* Take the freedesktop standard theme.
* Fallback to "our" icons.

jounathaen:
Curiously, in the Qt form designer, the Icons get shown correcty...

annejan:
I did read something about setting theme from environment a while back.

Interesting development :wink:

jounathaen:
Maybe it is a solution to set the Icons in the mainwindow.cpp and not in the mainwindow.ui ...

annejan:
Works on everything for me .  .
Arch, Fedora, Ubuntu . . 

Even nice dark themes :waning_crescent_moon: on both Gnome and KDE

jounathaen:
Ok, this seems to be a cinnamon specific Problem, I'll open an Issue

annejan:
I can now reproduce your cinnamon issue on a bare mint install.
Will look into the issue tomorrow

Issue #142, opened 2016-01-19
Bind a key to the clear action


ckundo:
May I suggest giving the clear button a keyboard binding, `escape`. This would make it quicker to operate with keyboard only, a good usability and accessibility enhancement. `escape` is conventional for exiting a modal state. Maybe a part of #84 .

annejan:
Done in `master`, will be in 1.1.1 release.

ckundo:
Thanks @annejan, great work!

Issue #85, opened 2015-08-04
All text input fields need example text & edit dialogue changes


lightonflux:
Just like this:

![bildschirmfoto vom 2015-08-04 02 19 08](https://cloud.githubusercontent.com/assets/1377943/9050555/438680f4-3a4f-11e5-94da-8218fae1324c.png)

----

I assume from the UI: the upper input field is the name of the things i create. And with the button next to it i confirm the action. The create thing then lands in box below.

![bildschirmfoto vom 2015-08-04 02 20 11](https://cloud.githubusercontent.com/assets/1377943/9050565/5aa33156-3a4f-11e5-882f-eadfb94160b3.png)

Okay i know the first is the password, the button is password generation and the last is the secret text input field, or what ever we want to call it. 

Suggestion:

 * *Add password*
 * *Gen.* – short for generate –, *Generate* or dice icon
 * *Write your secrets here*, *Fill me with secrets*, etc.

PS: Owncloud does not use a checkbox, but: 

![bildschirmfoto vom 2015-08-04 02 27 37](https://cloud.githubusercontent.com/assets/1377943/9050661/9398876c-3a50-11e5-83fb-da9238e96b66.png)

annejan:
The example "tips" in the txt fields are a great idea.

There is already some work done to add templating options to the add/edit screen in a separate branch.
For example: login and url
I think they should be displayed as a label + input field per item in the template.

In the `pass` format the password is always the first line, the rest is comments with optional key-value pairs with a colon to separate them.
That is why the password doesn't go to the main text field, since that is just for "comments".
But good to brainstorm a bit about the password field.

Is the "generate password button" really translated to: Erstellen?
That is almost as fun as "Eindrücken" for the git push button . .

lightonflux:
> Is the "generate password button" really translated to: Erstellen?

Yes. That a nice translation bug.

Also see [this beautiful translation](http://www.hyperblog.de/media/users/hoergen/bilder/2008/20081217_altweltraum.png).

Issue #84, opened 2015-08-03
Clear text input: use system icon instead of x


lightonflux:
I guess this just happened because something had to be in the button, and a label would be too big. Right?

It looks rather prototypy:

![bildschirmfoto vom 2015-08-04 00 58 36](https://cloud.githubusercontent.com/assets/1377943/9049590/f5509bc8-3a43-11e5-8462-3f060463be69.png)

Instead the system icon for "clear" should be used. Or better an inline x within the input box.¹  Or even more awesome and better: the inline clear icon of the system/toolkit.

Example (sorry, dark theme): 

![bildschirmfoto vom 2015-08-04 01 12 27](https://cloud.githubusercontent.com/assets/1377943/9049754/ffaecf52-3a45-11e5-87f4-13b652ce4a61.png)

1: as known from mobile websites and mobile safari

BTW: Just looked at my desktop apps (~120) and qtpass is the only one with a clear button

Oh look, you don't have to do it yourself: https://doc.qt.io/archives/qt-4.7/declarative-ui-components-searchbox.html

Other password managers like keepassx, keepass and last past choose another option. No clear functionality at all. 

*Source: quick google image search*

annejan:
Nice one . . Will have to find or design a button image that works on (just about) every theme.
Since apparently there is no system default button/widget available cross-platform.

lightonflux:
Mozilla does a good job with Firefox's icons. Just don't take VLC or keepass as examples. :)

annejan:
This is fixed in #145

PR #154, opened 2016-01-28
configwindow.ui default/start tab set to "settings"

jounathaen:
The settings tab is the first in order and also the probably most common used, so show it when opening the configwindow

### DIFF
--- a/configdialog.ui
+++ b/configdialog.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>561</width>
+    <width>564</width>
     <height>526</height>
    </rect>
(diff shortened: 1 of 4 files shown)


jounathaen:
OK, I wanted to make a new branch for the UI changes, but this somehow went wrong. Anyhow, some other improvements :-)

annejan:
Will check tomorrow and might do same (Escape key) thing with the User selection screen too.

annejan:
Solves: #142 and #84 and is a start of solving #85

annejan:
Hahaha . .we broke Qt 4.8 compatibility . . will fix in code (instead of .ui file)

PR #158, opened 2016-02-02
Two UI Tweaks

lftl:
These two changes probably aren't worth merging as is, but I thought someone might want to add config options for them and add them. I also know very little c++ and even less about QT, so the merges might be a horrible way of going about them.

1) When the window is activated, focus the search box. I generally leave QTPass open, and then just switch to it when I need a password. When I switch I always want to search for the password to copy. So for me it's really nice if search is always focused when I activate the window, and I can immediately start typing.

2) Usually when I've searched for a password it's the top item and I can just hit enter to grab it, but sometimes it's not and I need to navigate to another password. I set it up so the down arrow will move to the tree view below, but this can probably be solved more cleanly by simply adding the treeView to the tabstops after lineEdit.

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -897,6 +897,7 @@ void MainWindow::enableUiElements(bool state) {
   ui->updateButton->setEnabled(state);
   ui->treeView->setEnabled(state);
   ui->lineEdit->setEnabled(state);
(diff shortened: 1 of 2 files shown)


annejan:
Best commit all year :)
Hey, it's February..

PR #162, opened 2016-02-15
update gl_Es

xmgz:
I'm on it. Half job done


annejan:
:+1: will wait a bit and merge before next release.

Thanks for putting in the effort!

PR #168, opened 2016-03-06
Use --secure for pwgen and add more configurable options

ahippo:
Enable --secure option for pwgen by default as it allows it
to generate more random passwords.

Also, add the following options to pwgen together with corresponding checkboxes:
 * --capitalize
 * --numerals
 * --secure

All these options are enabled by default, as
 * we're a password manager, so no need to memorize passwords,
   so we should make them more secure by default;
 * most websites accept passwords, generated using these options
   (unlike --symbols option)

Convert horizontal layout for options into form layout
to accommodate more checkboxes.

### DIFF
--- a/configdialog.cpp
+++ b/configdialog.cpp
@@ -669,9 +669,13 @@ void ConfigDialog::setPwgenPath(QString pwgen) {
  * @brief ConfigDialog::on_checkBoxUsPwgen_clicked
  */
 void ConfigDialog::on_checkBoxUsePwgen_clicked() {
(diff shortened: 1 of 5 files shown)


annejan:
Wow, that is a nice change-set.
Thanks for your effort!
Giving it a quick run before I merge.

annejan:
Works great for me :)

ahippo:
> Wow, that is a nice change-set.
> Thanks for your effort!
> Giving it a quick run before I merge.
>
> Works great for me :)

Thank you very much for very quick response and merge!

PR #169, opened 2016-03-07
Remove path to password store in commit message and a leading space.

ahippo:


annejan:
Great work, thanks, way nicer way of handling these paths.

ahippo:
> Great work, thanks, way nicer way of handling these paths.

Thank you for merging this so quickly!

PR #175, opened 2016-05-03
Fix build issues with MSVC2015 on Windows

msvi:
Remove/Replace invalid compiler and linker flags
Do not include sys/time.h on Windows


annejan:
Thanks, that seems to check out on all non MSVC setups that are tested (automatically and manually).

PR #178, opened 2016-05-17
add Appdata file and update desktop file

daveol:
This adds an [appdata file](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html) for qtpass. The appdata file generally gets used by software centers to display things such as the description and screenshots.

This also adds the GenericName key to the desktop file and the language keys for both GenericName as well as Comment.

~~Things to get done before merge:~~
~~- [ ] Actually install the appdata file~~


daveol:
since the destop file doesn't get installed through qmake, I don't see the point in trying to install the appdata file.

PR #180, opened 2016-05-20
Prefer installed Qt5 instead of OSX's Qt4

dennisdegreef:
Qt5 appears to be installed, but not used in builds.


dennisdegreef:
Nevermind my derp, there are 4 builds and the last ones affect OSX.

barracuda156:
@dennisdegreef Does it still build with Qt4?

PR #195, opened 2016-09-04
Czech translation

svetlemodry:
for qtpass application


annejan:
Awesome, thank you very much for the effort!

svetlemodry:
@annejan: No probs, thanks for the quick merge!
I believe I also stumbled upon a typo in the english text - lines nr. 253+254 in `localization_en_GB.ts` file.
`... dialouge ...` => `... dialogue ...`

annejan:
Thanks . . 
Fixed those too 😸

annejan:
It's a typo in the actual code, copied to the translation.
Added a en_US too for those afflicted, since fixing it in code would require all translations to be fixed, that's something for another day.

Issue #101, opened 2015-09-23
Character encoding issue with GPG key


jdhoek:
In the *Users* dialog I am encoutering some character encoding issues:

![encoding](https://cloud.githubusercontent.com/assets/683699/10044600/951795fa-61fc-11e5-9c93-38125c4292ec.png)

The public key in the screenshot belongs to someone whose name ends with `qué`, but in the user interface this is rendered as `quÃ©`. This usually means that UTF-8 text is decoded incorrectly as ISO-8859-1 text.

This public key causes no problems in Gnome's keyring or in the command line gpg tool.

My locale is all in UTF-8:

```
LANG=en_GB.UTF-8
LANGUAGE=en_GB:en
LC_CTYPE="en_GB.UTF-8"
LC_NUMERIC=nl_NL.UTF-8
LC_TIME=nl_NL.UTF-8
LC_COLLATE="en_GB.UTF-8"
LC_MONETARY=nl_NL.UTF-8
LC_MESSAGES="en_GB.UTF-8"
LC_PAPER=nl_NL.UTF-8
LC_NAME=nl_NL.UTF-8
LC_ADDRESS=nl_NL.UTF-8
LC_TELEPHONE=nl_NL.UTF-8
LC_MEASUREMENT=nl_NL.UTF-8
LC_IDENTIFICATION=nl_NL.UTF-8
```

jdhoek:
QtPass version 1.0.1.

annejan:
Related: https://github.com/IJHack/qtpass/issues/91 encoding issues opening or writing a password.

annejan:
Fixed in #198

PR #198, opened 2016-09-24
issue 101 bugfix

asalamon74:
I had the character encoding problem described in issue #101 
This patch fixes it.


annejan:
Nice find!
Never used QTextCodec before.
Seems like it might be useful in other places too.

Thank you for the bug fix!

Issue #192, opened 2016-07-31
German umlauts fails


haukehein:
In the encrypted data  QtPass 1.1.3 shows the German umlauts (Ä,ä,Ö,ö,Ü,ü,ß) always wrong. It shows about instead of   ö   to diie mark   Ã¶     .

on: AMD64 PC with Debian, 64 bit,

annejan:
Closed in #199

Issue #131, opened 2016-01-01
Bug: Special characters in Template


jounathaen:
When I create a template field such as "Prüfnummer", it gets displayed as something like "PrÂ½fnummer".

annejan:
Will have to look into utf8 encoding/decoding it seems.

annejan:
Closed with #199

Issue #91, opened 2015-08-27
saved password '§' turns to 'Â§' when copied to clipboard or shown when editing


pittagurneyi:
Hi, 

thank you for this excellent software.

Sadly, I found a rather problematic error. When saving a password with § in it, the actions of copy to clipboard or showing the password when editing, turn it into Â§, which alters the password.

If someone could perhaps test it, to make sure that the issue is not solely on my system, that would be great.

Using pass on the console returns the correct password. Should there in fact be an issue with qtpass, it would be really helpful to get a list of characters, that exhibit this sort of problem, to verify all existing passwords against possible corruption.

Thank you!

annejan:
Purely off the top of my head this looks like an **encoding** issue.
I'm pretty sure the clipboard either expects UTF-8 and is presented an ISO-8859 string or vice-versa.

Would you mind telling me which platform (and/or window manager) you are using?

pittagurneyi:
Gentoo Linux, KDE Plasma5

The problem is, that, upon clicking "show password" (or whatever it says in english), in the edit dialog, it also shows the password as Â§.

annejan:
Thanks, I run the same (well, Arch) on my main laptop, so that should make debugging pretty easy.

The fact it also shows the password corrupted makes it even easier . . 
Apparently the process that fetches the password already makes the encoding mistake.
Will try and look into this as soon as possible, thanks for finding and the clear description!

pittagurneyi:
Thank you for looking into it!

If you could notify me when you have the patch ready, I'd like to fix this on my system as soon as possible.

pittagurneyi:
It looks to me like process->readAllStandardOutput is not properly read as QString utf8.

Perhaps QString::fromUtf8 would help.

pittagurneyi:
I just tested it and the password is received okay. Just plain § now.

Just made sure all process->readAllStandardOutput is properly wrapped in QString::fromUtf8 in mainwindows.cpp

Since I don't know the whole picture of how the program is written and don't have the time to work through it all, you have to make sure this actually works and doesn't create problems elsewhere.

I'll look forward to hearing from you.

annejan:
Wow, thanks for the effort!!

Will do some more testing (on Linux, Mac and Windows) tonight . . 
Yesterday only had time to do a quick test on OSX and there it worked as-is . . so can easily do regression-testing..

pittagurneyi:
[...]

PR #199, opened 2016-09-26
issue 91 bugfix

asalamon74:
Yet another character encoding bugfix. It solves issue #91. Also solves #131 and #192.


annejan:
Thank you very much for this effort @asalamon74

PR #200, opened 2016-09-27
Fix spelling error

innir:
Fix trivial spelling error in `configdialog.cpp`.


annejan:
Thanks for this find.

PR #201, opened 2016-09-27
Lookup validity field to check if keys are valid

thotypous:
Currently, the code checks the Ownertrust field [1] to look if a key should be considered as valid. However, the Ownertrust field should not be deemed to represent the level of confidence that a key is valid. Rather, it represents how much the user trusts in the owner of the key to understand correctly how key signing works and to strictly check fingerprints before signing keys. Ownertrust is used as input by the trust models to compute the validity of keys, which is printed by GnuPG in the Validity field.

This commit changes the code to check the Validity field instead of the Ownertrust field.

Keys which are at least marginally valid are also included in the user list dialog, however keys which are not fully valid are printed with dark orange background.

[1] http://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob_plain;f=doc/DETAILS

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -1380,7 +1380,7 @@ QList<UserInfo> MainWindow::listKeys(QString keystring, bool secret) {
       current_user = UserInfo();
       current_user.key_id = props[4];
       current_user.name = props[9].toUtf8();
-      current_user.validity = props[8][0].toLatin1();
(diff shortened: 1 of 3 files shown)


annejan:
This indeed seems to be a more valid way to handle the key trust.
I'll have to check if this is also the way `pass` handles key trust. I sort of remember a discission about this.

Merging to master, thank you very much for your contribution.

Issue #202, opened 2016-09-30
Short fullname hangs QtPass keypair generation process for infinite time


dakk:
I was trying to initialize QPass by creating a keypair. I put my fullname and other requested info. Then the "Generating a gnupg keypair" window appears, but after 3 hours it never finished.

So I tried to create a keypair from terminal using the same backend used by QPass (gpg2); when I type my fullname (I typed "dakk" that's my nickname) gpg2 says that the fullname should be longer than 4 characters. 

I figured out that QPass received this error by gpg2, but this error is not handled properly by QPass.

> gpg (GnuPG) 2.0.30
> libgcrypt 1.7.3
> Home: ~/.gnupg
> Supported algorithms:
> Pubkey: RSA, RSA, RSA, ELG, DSA
> Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
>        CAMELLIA128, CAMELLIA192, CAMELLIA256
> Hash: MD5, SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
> Compression: Uncompressed, ZIP, ZLIB, BZIP2

dakk:
I forgot to mention my gnugp version:

> gpg (GnuPG) 2.0.30
> libgcrypt 1.7.3
> Home: ~/.gnupg
> Supported algorithms:
> Pubkey: RSA, RSA, RSA, ELG, DSA
> Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
>        CAMELLIA128, CAMELLIA192, CAMELLIA256
> Hash: MD5, SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
> Compression: Uncompressed, ZIP, ZLIB, BZIP2

egilh:
First time qtpass user here.
I can confirm I have the same issue.
Trying to create a keypair, filling in my email and full name, it hangs on "Generating a new keypair".

Adding a letter to my given name makes it possible to create a keypair. Doesn't matter if I also use my surname or not, its the length of the given name that creates the issue.

annejan:
Solution seems to be two-fold.

A) Correctly check for status / errors of gpg2 keygen.
B) Warn that the fullname should be longer than 4 characters.

dakk:
I implemented your (B) solution. I also check for email validity (gpg2 expect for a valid email address)

PR #203, opened 2016-10-03
keygendialog email and name validition (issue 202)

dakk:
This pull will solve #202. Now when you press the generate button in keygendialog, the software checks for email (with a regexp) and name validity. If something is wrong, it display a QMessageBox::critical


dakk:
That was a pleasure for me (y)

Issue #98, opened 2015-09-11
More options for password generation


sjau:
There are still some websites out there that have very limited options when making a password. Believe it or not, some websites still won't allow more than 12 characters. Others might now allow special chars etc.

IMHO it would be good to be able to give some input on how the password should be generated, e.g.

min/max length
whether to include/require:  a-z, A-Z, 0-9, special chars

I think LastPass with it's advanced form makes already a very good generator: https://neoblog.lastpass.com/wp-content/uploads/generatepassword.png

annejan:
QtPass has plenty options builtin . . 
Both with and without `pwgen` selected . .

If you have a good idea about using a different (kind) of password generator, please re-open this issue.
These options are indeed not usable from the password creation or editing screen, if you really think they should be, feel free to re-open this issue too.

jounathaen:
I think it would be nice to choose the Password options in the add password dialouge.
I often generate a password, enter it on the website, then the website complains about special characters or length. Now I have to close the add password dialouge, go to the options and after that, open the new entry dialouge again. So I think, there should be options on the password generation in the editing screen.

jounathaen:
I'm currently working on it

annejan:
Needs a bit more work, I have [x] Use `pwgen` option enabled and the drop-down selection doesn't affect the password generation . .
Great start though 👍

annejan:
The passwordDialog template stuff seems to have suffered some regression too.
<img width="609" alt="screen shot 2016-10-12 at 22 03 05" src="https://cloud.githubusercontent.com/assets/294470/19325749/daafa6ca-90c7-11e6-86a9-ec0f99b65cca.png">

annejan:
Disabled character selection when using `pwgen` . .
Might swap out the dropdown for pwgen options at a later date.

PR #207, opened 2016-10-11
Password templates

jounathaen:
Added password character sets.
Three default and one customizable.
In the password generation dialouge, you can now choose the lenght of the generated Password and the character set to choose the characters from.
Should solve #98 or at least make it a lot better.
Greetings


annejan:
Thank you very much for this useful addition,  going to have to play with it a bit but looks great sofar.

PR #209, opened 2016-10-17
Increase maximum password length to 255

vladimiroff:
QSpinBox has a default maximum of 99 thus making it impossible to
generate longer passwords.

http://doc.qt.io/qt-5/qspinbox.html#maximum-prop


annejan:
Thank you very much for this addition.

I'll also add this change to the (new) QSpinBox in the password add/edit dialog.

PR #211, opened 2016-10-19
Allow ssh links

cgonzalez:
In most Linux desktops, it is allowed to use ssh links (ssh://user@server) but QtPass only allow http and ftp links.

Tested on KDE Plasma 5.


annejan:
Thank you very much, this should also works on macOS . .
And apparently can be made to work on windows too http://superuser.com/questions/204066/make-ssh-links-open-with-putty

PR #216, opened 2016-11-13
creating password store directory, if it doesnot exists, de_DE translation fixes and removed obsolete translations

YoshiMan:


annejan:
The more important bit is https://github.com/IJHack/QtPass/pull/216/commits/d93aa124f895527e5ff493b4a8e366daeee6f055 creating password store directory, if it doesnot exists.

Seems like a good addition to the "first run" scenario and makes adding profiles easier.

YoshiMan:
I really struggled because i didnt had the *$USERHOME/.password-store* directory. So on the right side of the mainwindow was an info *QProcess::FailedToStart* and i had no idea, whats wrong. The working directory of the was process ($USERHOME/.password-store) didnt exist. That was the fault.

YoshiMan:
thanks for merging
yes, the title is better ;)

PR #218, opened 2016-11-16
added a copy button for each line to paste the content into the clipboard, "pass init -- path=" command with right path-parameter, lupdate qtpass.pro

YoshiMan:
Now we have the possibility to copy the additional-information from the password-file into the clipboard.
Maybe thats not the right QT-way, feel free to correct this. Im new on QT and it works. :)
See attached screenshot. e.g on doubleclick will the username or the url be pasted into the clipboard and be cleared after the amount of the configured autoclear.
enabledoubleclicktocopyintoclipboard

The `pass init` command, which was executed, when you add new users to the password files, was executed full qualified. e.g. `pass init --path=/home/user/mypassstore/DBpasswords ...` , and the password store is locate under **/home/user/mypassstore**. This would result into a new folder into the password storage **/home/user/mypassstore/home/user/mypassstore/DBpasswords**, because `pass init --path=...` sees its `path-parameter` relativ to the passstore.
see mainwindow.cpp:1485
unfortunately no extra commit :( sry

I also executed the `lupdate qtpass.pro` command to re-generate the *.ts files. Now they are in sync with the code again.


jounathaen:
Nice, I wanted to have something simmilar.
Suggestion: Don't copy on doubleclick on the lable, but add a small copy button at the end of each line.
I think this would be easier to understand for other users.

YoshiMan:
Would be a nice idea too.
But the, we should move the "copy password"-button down there too, right?
And i would place at the begining of the line, not the end. So its closer to the left panel, where you last clicked.

jounathaen:
Totally agree!

YoshiMan:
okay. hopefully i will be able to push it today.

YoshiMan:
so i pushed it, but the button image is just the **:/icons/go-top.svg**. Please change it :)
And please heavy test this feature ;)

jounathaen:
Can you upload a screenshot? I can currently not merge?

YoshiMan:
I'll upload it tomorrow morning. Just found the way to my bed ;) its 00:10 in germany :P

YoshiMan:
Good morning (for me). Here are the screenshots.
The copy password button is gone, but each line got its own one. (of course still the wrong image).

**showpassword notemplate showcontent**
showpasswordnotemplateshowcontent
**showpassword usetemplate showcontent**
showpasswordusetemplateshowcontent
**hidecontent**
hidecontent
**hidepassword usetemplate showcontent**
hidepasswordusetemplateshowcontent
**hidepassword notemplate showcontent**
hidepasswordnotemplateshowcontent

annejan:
Good morning . .
Awesome work @YoshiMan 

Would you like me to merge this pull request and fix the icons and check on all behaviours . .
Or would you like to continue with this pull request some more?

YoshiMan:
Pls merge and heavy test :)
We can open new issus, if have some more work with this feature.

Issue #127, opened 2015-12-30
Cannot create top level folder


muff1nman:
I cannot create folders at the root `/`. Right clicking a folder lets me create a subdirectory and right clicking a top level file does not give an option to create a folder.

Looking through previously closed issues, it looks like one is supposed to be able to create top level folders by clicking in an empty white space. However, I have no empty white space as the window is filled with existing folders/files and has a scroll bar.

Suggestion: Create a new button to create a folder.

Thanks and awesome work.

jounathaen:
Also got this problem. You simply cannot select the root directory by clicking on empty space in the password list.

annejan:
This must be some sort of regression, as clicking empty space was supposed to select the root folder. Will have a look as soon as I can.

muff1nman:
To clarify, clicking on the empty space works. The problem arises if there
is no empty space due to the amount of entries in the list.

bmiddleb:
I don't see that it works regardless of whether there's empty space or not.

annejan:
Perhaps "empty space" was a bit ambiguous @bmiddleb 

Right mouse button in empty spot in tree view:
<img width="246" alt="screen shot 2016-06-20 at 21 15 57" src="https://cloud.githubusercontent.com/assets/294470/16207245/5ab08bf0-372c-11e6-8f6c-e285ef2b9856.png">
Select top option:
<img width="437" alt="screen shot 2016-06-20 at 21 16 10" src="https://cloud.githubusercontent.com/assets/294470/16207258/6aea7e5e-372c-11e6-80bf-b4981ebb3247.png">

If there is not enough empty space in te tree view for a right click, a simple solution is searching for some gibberish in the search box above the tree view.

sanbor:
I had to find this issue to see how to create a dir at the top level. I suggest adding `.password_store` or the root directory in the tree. This way it would be very intuitive how to select the top folder and add a new child.

annejan:
Should be solved by #220

khanson679:
I'm on 1.1.6 on Kubuntu 17.04. Still can't deselect folders in the tree view by left clicking on whitespace. Oddly, right clicking deselects (in addition to bring up the menu option that works).

PR #220, opened 2016-11-17
UI Improvements

jounathaen:
So, as announced, i did some work on the UI.
you can now deselect items in the TreeView and the Mainwindow looks a bit nicer. Added Icon in corner


jounathaen:
Solves #127

PR #222, opened 2016-11-17
Moved @YoshiMan 's copy buttons inside the line Edit

jounathaen:
Think this looks a bit more elegant (at least on Cinnamon)
qtpass1

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -2066,18 +2066,33 @@ void MainWindow::addToGridLayout(int position, const QString &field, const QStri
 {
     QString trimmedField = field.trimmed();
     QString trimmedValue = value.trimmed();
+
     QLineEdit *line = new QLineEdit();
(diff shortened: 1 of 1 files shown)


jounathaen:
I can't test these layout changes on other window toolkits, so i'd be very interested, how this looks on Windows, Mac, KDE, etc.
Screenshots would be great :+1:

annejan:
screen shot 2016-11-17 at 20 06 26

YoshiMan:
My Screenshot from Antergos Linux (4.8.8-1) and Gnome (3.22.0-1)

antergoslinuxgnome

annejan:
I think I'll merge this (in the morning) and have a look at swapping the copy button to the right.

YoshiMan:
i think its not a good idea to move the button to the right, because you select a password on the left panel. If you place the button on the right you have to move the mouse across the whole screen.

PR #224, opened 2016-11-20
refactoring, new QtPassSettings class, all settings should be read and written here

YoshiMan:
I refactored the mainwindow.
Now we have a QtPassSeetings class, where you can read the settings in a static way.
e.g. QtPassSettings::getVersion(), you will get the version as int or "0" as default fallback 
or QtPassSettings::getVersion(4), you will get version as int or "4" as fallback

hope you like it.
can you please check for memory leaks or other mistakes i made.


annejan:
Thanks for this big change of the configuration model.
This makes a lot of things way more readable and probably stable too.

I'll boot up a Windows VM and fix the windows errors.
After that I'll do some thorough testing and checking.

YoshiMan:
yeah that was a big change. took me some time ^^
Additional checks should be made through version changes. especially the **useClipboard** config.
If you fresh install, everything should work. But i dont know what happens, when you have an older version.

YoshiMan:
@annejan. how was your testing?

annejan:
So far so good, both "scenario driven" testing and static analysis haven't found any regression.

I was also triggered to do some other minor fixes in other places that I had postponed out of lack of inspiration.

Great work @YoshiMan thanks for the effort!

YoshiMan:
Your welcome. QtPass is a realy nice tool.
What is "senario driven" testing?
And can we do some thing like unit-testing?
I would appreciate some unit-tests. :)

annejan:
Scenario driven in this case is just a lot of different configs I can easily switch out and a big set of different password files (some broken on purpose) . .

I would love to add unit testing and have looked into it a bit a couple of months ago but could not find a good starting point.
I've been working test driven on a couple of projects recently and really like the way of working from idea to test to code and seeing the tests clear one by one . .

Qt has a framework for unit tests http://doc.qt.io/qt-5/qtest-overview.html although I have to admit I have no experience with it yet.
If I can find the time the coming days I'll try and set up some minimal tests and hook it up in the CI flow of the project.

YoshiMan:
Alright, so you are doing some "manually unit-testing" :D
I'll have look at the testing framework, too.
I don't have any experience with cpp or Qt. So it is important that someone reviews my code and that there are some unit tests :)

PR #234, opened 2016-11-23
refactoring - pass ifce, process mgmt

tezeb:
This is a first refactoring step to extract functionalities from MainWindow. It compiles and works(mostly?). I'm sharing to propose direction and gather feedback. It consist of 3 commits.
First one(e15c93f56640ef6810d3304a07fd1e0fd4bfe350) is an interface _"design"_ for _"backend"_(pass or (git/gpg...)). This was my main driver for refactoring.
Second commit(1e06efd1bdfc55a4d03d454137256b46e4389b26) moves code from MainWindow to appropriate classes. I've tried keeping changes to minimum to ease reviewing at this stage. Unfortunately, as it showed up, I had to move (probably all) process management code into my interface as well. I think it shall be further separated.
Last commit contains localization updates.

As I've already mentioned next step is extraction of process mgmt into separate class. I also think that there is a huge potential for using signal/slots mechanism for MainWindow<->Pass Interface(process mgmt) communication.

And special thx goes to @YoshiMan for extracting QtPassSettings, which was a huge help!

### DIFF
--- a/imitatepass.cpp
+++ b/imitatepass.cpp
@@ -1,0 +1,286 @@
+#include "imitatepass.h"
+#include "mainwindow.h"
+#include "qtpasssettings.h"
+
+ImitatePass::ImitatePass()
+{
+
+}
+
+/**
+ * @brief ImitatePass::GitInit git init wrapper
+ */
+void ImitatePass::GitInit() {
+    executeWrapper(QtPassSettings::getGitExecutable(),
+                   "init \"" + QtPassSettings::getPassStore() + '"');
+}
+
+/**
+ * @brief ImitatePass::GitPull git init wrapper
+ */
+void ImitatePass::GitPull() {
+    executeWrapper(QtPassSettings::getGitExecutable(), "pull");
+}
+
+/**
+ * @brief ImitatePass::GitPush git init wrapper
+ */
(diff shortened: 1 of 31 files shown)


annejan:
This sounds like exactly the kind of refactoring I was thinking about but never got the focus to do :)
Thank you very much, will look into the changes and provide feedback tomorrow.

jounathaen:
Great, that somebody finally starts this :-)
 👍 👍 👍

annejan:
Moved the `#ifdef __APPLE__`

In the not so wise words of me 😈 
Works on my machine .. merged!

annejan:
According to static analysis this change even removes a identical_branches "issue" in a moc_
https://scan.coverity.com/projects/ijhack-qtpass and reduces complexity a lot . . Thanks for the effort @tezeb !!

YoshiMan:
@tezeb awesome work. I love it. thumbs up

tezeb:
wow, that was fast merge :) thx
Glad that you liked the changes.

@annejan regarding "housekeeping" commit, do you have some tool/config for fixing style?

annejan:
I have set up clang default standards in vim and QtCreator . .
To fix in commandline I use `clang-format -i *.cpp *.h`

PR #242, opened 2016-11-27
Password dialog decoupling from MW

tezeb:
Another dependency removed. Also generatePassword gone from MW.
I've added new header "datahelpers.h", which might not be a good name, but it contains more widely used structures(two of them so far).

Can we rename localization to xlocalization, so that they will render last here? ;)


annejan:
Awesome work!

tezeb:
thx. I haven't worked with Qt in a while, but it's so much fun :D

PR #249, opened 2016-12-05
Process management refactoring part 2

tezeb:
Further extraction of process managment. It seems to work both with pass and natively. Tested only on Linux though. Mostly moving code around, but also added new interfaces.
As previously I tried to split it into logical(somehow) parts.
1. 6353745 Initial interface of process executor
2. c464f6e I allowed myself to add some more info to debug logs(in ugliest possible way, sorry), but it shows file/linenumber now, which speeds up looking for debugs/issues.
3. aa0eb00 This is ready executor, which I am using in half baked part 3 of refactoring(with few minor changes).
4. b11e71d Switch everything possible to use new executor, to hide PM from application.
5. lang update ofc

Next steps:
1. change MW<->Pass communication to multiple signals (for each action type(see PROCESS enum inside Pass)).
2. add "transactions" - I've noticed that especially with native programs, if there is an error with first execution, the error is hidden from view, as it's replaced with results of further executions. It's hard to reproduce and probably does not happen very often, but it will allow for better error handling. What do you think?

### DIFF
--- a/configdialog.cpp
+++ b/configdialog.cpp
@@ -3,7 +3,7 @@
 #include "mainwindow.h"
 #include "qtpasssettings.h"
 #include "ui_configdialog.h"
-#include <QDebug>
+#include "debughelper.h"
 #include <QDir>
 #include <QMessageBox>
 #ifdef Q_OS_WIN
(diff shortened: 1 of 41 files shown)


annejan:
Awesome work .. this will make debugging some of the current issues a lot easier 👍

Issue #250, opened 2016-12-05
Pass environment not set-up correctly


annejan:
After merging #249 using `pass` with a non-standard PASSWORD_STORE_DIR doesn't seem to work correctly . . 
Opening those passwords does work in "native" mode . .

Related issue #248

tezeb:
#248 is not related(it's present in 1.1.5 as well).

Fixed in https://github.com/IJHack/QtPass/pull/251.

PR #251, opened 2016-12-05
Minor fix for file names and git push

tezeb:
Fix for refactoring bug, was adding file with extra '.gpg' suffix. This also clarifies a bit Pass interface, when passing filename it requires to be in pass format(ie. password-store relative path and no '.gpg' suffix). It is resolved internally(for native tools).
Fix for pushing to git, when "Auto push" is checked but "Use git" is not.


annejan:
Fix for pushing to git, when "Auto push" is checked but "Use git" is not.
Should fix: #250

Issue #243, opened 2016-11-28
Double-clicking might open previous entry instead of one double-clicked on


annejan:
* Click first entry, wait for it to open
* Double-click on second entry

> **Current behaviour**
First entry data is opened for editing (second entry name is shown as location in dialog title)
When selecting [OK] second entry data is overwritten with (edited) content of first entry 
*Second entry is shown with overwritten data*
When selecting [Cancel] second entry is shown 

> **Expected behaviour**
Second entry is opened for editing

Related to #230

YoshiMan:
@annejan Could you assign it to me?

YoshiMan:
this should be fixed, right?

annejan:
It should, will test after dinner :)

PR #254, opened 2016-12-11
executeing pass show before editpassword dialog shows up

YoshiMan:
Double-clicking might open previous entry instead of one double-clicked on #243

### DIFF
--- a/executor.cpp
+++ b/executor.cpp
@@ -140,9 +140,9 @@ int Executor::executeBlocking(QString app, const QStringList &args,
     QTextCodec *codec = QTextCodec::codecForLocale();
     QString pout = codec->toUnicode(internal.readAllStandardOutput());
     QString perr = codec->toUnicode(internal.readAllStandardError());
-    if (process_out != nullptr)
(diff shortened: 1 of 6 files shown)


YoshiMan:
While working on this ticket, this `lastDecrypt` property was very confusing for me. Do we really need it?
Should it be removed? I think we could handle this with the signal/slot concept?

annejan:
I think lastDecrypt is only used to check weather or not the clipboard should be cleared . .

We don't want to clear the clipboard if it doesn't contain the password (or something else from our app) but the user has copied something else . .

YoshiMan:
hm. i think its the `MainWindow::clippedText`, you are talking about. This one handles the clipboard related stuff.

annejan:
Yep .. so we could/should probably completely refactor out the lastDecrypt stuff . .

That's still a remnant from when QtPass was a simple read-only `pass` viewer for management ;)

YoshiMan:
Alright. I'll open an Issue.

Issue #258, opened 2016-12-14
Feature: CTRL/CMD + Q closes the mainwindow


YoshiMan:
As a user of QtPass
i want to quit the application with a keyboard shortcut,
so that i dont have to use the mouse on move it all round the screen
and save a little bit time ;).

You already can close dialogs with ESC.

**Given** the mainwindow is focused
**When** CTRL/CMD+Q is pressed
**Then** the QtPass application closes.

**Given** any dialog is focused
**When** CTRL/CMD+Q is pressed
**Then** nothing happens.

PR #259, opened 2016-12-14
Feature: CTRL/CMD + Q closes the mainwindow #258

YoshiMan:
enabled shortcut
extra treatment for mc os

### DIFF
--- a/mainwindow.cpp
+++ b/mainwindow.cpp
@@ -9,7 +9,7 @@
 #include <QQueue>
 #include <QTextCodec>
 #include <QTimer>
-
+#include <QShortcut>
 #ifdef Q_OS_WIN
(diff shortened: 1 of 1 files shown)


YoshiMan:
please test under Mac OS

annejan:
Travis-CI will fail on macOS when it finally gets to that @YoshiMan  ;)

http://doc.qt.io/qt-5/qkeysequence.html#qt_set_sequence_auto_mnemonic
Note: This function is not declared in any of Qt's header files.
To use it in your application, declare the function prototype before calling it.

But it is available via <QtGlobal> on 4.8 . . 

I could merge and fix if you want me to . .

YoshiMan:
Yeah. It would be better, if you could fix this. Thank you.

Issue #239, opened 2016-11-24
drag and drop for passwords and passwordfolders to move


YoshiMan:
### moving a folder into a folder
1. @Given The user enabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview a folder1 onto folder2.   
3. @Then Folder1 would be moved into folder2 via `pass mv folder1 folder2/folder1`

### moving a folder into a file
1. @Given The user enabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview folder1 onto file1.   
3. @Then nothing will happen.

### moving a file into a file
1. @Given The user enabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview file1 onto file2.   
3. @Then nothing will happen.
(we could think of `pass mv -f file1 file2` , but this would force an overwrite)
@annejan what do you think?

### moving a file into a folder
1. @Given The user enabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview a file1 onto folder1.   
3. @Then File1 would be moved into folder1 via `git mv file1 folder1`

### moving a folder into a folder
1. @Given The user enabled the *git* setting, but disabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview a folder1 onto folder2.   
3. @Then Folder1 would be moved into folder2 via `git mv folder1 folder2/folder1`

### moving a folder into a file
1. @Given The user enabled the *git* setting, but disabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview folder1 onto file1.   
3. @Then nothing will happen.

### moving a file into a file
1. @Given The user enabled the *git* setting, but disabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview file1 onto file2.   
3. @Then nothing will happen.
(we could think of `git mv -f file1 file2` , but this would force an overwrite)
@annejan what do you think?

### moving a file into a folder
1. @Given The user enabled the *git* setting, but disabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview a file1 onto folder1.   
3. @Then File1 would be moved into folder1 via `git mv file1 folder1`

### moving a folder into a folder
1. @Given The user disabled the *git* setting and disabled the *pass* setting.
2. @When The user drags and drops within the passwordstoragetreeview a folder1 onto folder2.   
[...]

PR #261, opened 2016-12-15
#239 reencrypting after a drag and drop action

YoshiMan:
drag and drop for passwords and passwordfolders to move passwords and passwordfolders 
reencrypting after a drag and drop action


YoshiMan:
@annejan 
I hope this solution is fine for you.

PR #270, opened 2017-01-08
Process specific signals for process management

tezeb:
Added process specific signals for easier process management. Also a bit of cleanup and (automatic) update of coding style in few places.


tezeb:
This also fixes Real/Imitate-Pass signal/slot connection that was recently broken.

Issue #262, opened 2016-12-15
Bugs since refactoring


jounathaen:
Hey,
Unfortunately, since the refactoring, I got several Bugs:

- qmake && make && sudo make install does not work
```
cd src/ && ( test -e Makefile || /usr/lib/x86_64-linux-gnu/qt5/bin/qmake /home/foo/src/src.pro -o Makefile ) && make -f Makefile install
make[1]: Entering directory '/home/foo/qtpass_official/src'
make[1]: Nothing to be done for 'install'.
make[1]: Leaving directory '/home/foo/qtpass_official/src'
```

- No more Copy Icons
They are simply missing, when starting the Program from QtCreator
![qtpass1234](https://cloud.githubusercontent.com/assets/16337548/21237527/4ccca1ca-c2f7-11e6-928f-9b45f531b78b.png)

- Creating Passwords with pwgen does not work
Use pwgen is enabled, when clicking on generate, nothing happens.
Commandline says  `pass.cpp : 84 	 pwgen fail` When unchecking "use pwgen" it works, but since #238 this should not be used... (pwgen is installed in Version 2.07)

As Always: Linux Mint - QtCreator 3.5.1 - Qt 5.5.1

Greets!

annejan:
It might be that a make clean doesn't work after the move, the broken make install might have to do with that . . haven't tried the install after the change (just the deploy to package routes) to be honest . .
The `pwgen` issue I can confirm . . on both Linux and macOS

Thanks for noticing . . and doing this writeup . .

Will have a look tomorrow 😄

annejan:
I know it sounds dangerous but `rm *.cpp *.h *.o` in the main qtpass folder might help . .

tezeb:
Regarding the copy icons, it might be because of move into src directory. When I tried viewing `mainwindow.ui` in qtcreator it asked for location of `resource.qrc' file, which was then added into `qtppass.pro`. It's commited since yesterday (https://github.com/IJHack/QtPass/pull/270), so it may already work.

tezeb:
Regarding `pwgen`, the problem is with spaces at the end of arguments. I'll fix this soon.

tezeb:
Only `make install` left. It seems that `Makefile` generated in `src/` has empty `install` target. @annejan any ideas?

annejan:
Not really, @tezeb but will have a look later today . . 

Currently on a mac so not using make install at-all . . will check on my Arch machine tonight.
[...]

PR #271, opened 2017-01-09
Fix pwgen and refactor Pass::finished

tezeb:
Fix for https://github.com/IJHack/QtPass/issues/262.
Also "revert' fix for break added in switch. It was unreadable, sorry for that. I hope it's better now.


tezeb:
On the bright side you made me update it ;)

PR #275, opened 2017-01-09
Final step in process mgmt refactoring

tezeb:
It's an extension to fix from one of my commits in previous pull request.
I properly hid password, but as it showed up, other fields where still visible duplicated when showing template fields.

### DIFF
--- a/src/enums.h
+++ b/src/enums.h
@@ -5,19 +5,34 @@
     \class Enums
     \brief Enumerators for configuration items
 */
-class Enums {
-public:
-  /**
-   * @brief Enums::clipBoardType enum
-   * 0 Never
-   * 1 Always
-   * 2 On demand
-   */
-  enum clipBoardType {
-    CLIPBOARD_NEVER = 0,
-    CLIPBOARD_ALWAYS = 1,
-    CLIPBOARD_ON_DEMAND = 2
-  };
+namespace Enums {
+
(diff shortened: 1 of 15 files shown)


tezeb:
It was supposed to be separate pull request. But as there is not that many changes, it can go as one.

So, this is final step in refactoring process management(except for unit tests and bugfixing).
The idea is that normal `pass` behaviour is what we want from `ImitatePass`, thus I've added transaction mechanism inside it. Now MainWindow is signaled only when `pass`-like action is complete. Previously all processes where signaling MainWindow, which made QtPass behave differently when using `RealPass` vs `ImitatePass`. All comunication is also done via dedicated signals(one signal per `pass`-like action), which makes it easy to extend MainWindow reactions.
From now on any differences in behaviour(when using different backends) shall be treated as a bug. Also it is assumed that any file names passed to `Pass` interface are in `pass` style, as MainWindow shall not care(nor check) which backend is in use.
I plan on adding some unit tests, to verify that both backends have exactly the same behaviour/interactions/signaling.

tezeb:
I skipped lupdate to ease code review.

annejan:
I've also ran these changes through coverity static code analyser
https://scan.coverity.com/projects/ijhack-qtpass

Thanks for the great work (again) @tezeb !

PR #276, opened 2017-01-16
small band aid fix for password generation on windows

treat1:
Uses BCryptGenRandom win32 api function but still falls back to qrand. A proper fix would require more changes further up the stack maybe a dialog with retry and cancel button if password generation should fail. Could also be extended to use other apis on other platforms e.g. /dev/random on linux. Does this sound worthwhile or is it being work on elsewhere?


annejan:
Aparently bcrypt is not available on Windows as-per standard.

```
fatal error: bcrypt.h: No such file or directory
 #include <bcrypt.h>
```

Any idea how to install it from within `.appveyor.yml` ?

treat1:
It's an issue with mingw in appveyor then. Bcrypt is included with windows since vista. I have the latest mingw from qt installer installed and it has the bcrypt.h/lib included.

annejan:
I'm downloading a Windows VM from modern.ie to check this one out :)

Would like to have a working appveyor (since it does the windows builds so I don't have to) before merging.

treat1:
The header file doesn't come bundled with windows only the dll. The header file comes from mingw. Looking at the log it seems to be using the same version of mingw thou 5.3 maybe it's not picking up the correct paths somehow.

treat1:
On my computer It's found in C:\Qt\Tools\mingw530_32\i686-w64-mingw32\include\bcrypt.h

treat1:
If you can't get it working a work around would be to use LoadLibrary/GetProcAddress to load the library instead.

treat1:
Possible fix. The path on your appveyor thing should include C:\Qt\Tools\mingw530_32\bin not C:\MinGW\bin

annejan:
That did the trick! 
Thanks for this update @treat1, I'll look into the /dev/random and other implements later on :)

PR #304, opened 2017-03-12
Code coverage measurement

### LINKED ISSUE #298: Measure unit-test code coverage
As unit tests were recently added, it would be a nice idea to also add code coverage.
It seems as travis is able to gather gcov data, which should make it relatively easy to setup.
I'm willing to set it up(at least for Linux), although I am not a travis expert.

But, as there are also alternatives, I would like to discuss it briefly.
What are your thought?

tezeb:
We now have an ability to measure code coverage(stunning 2% right now ;)).  This closes #298.

While the commit may seem a bit complicated at first, everything boils down to compilation with `--coverage`. Basically this have to be a new 'flavour' of build as coverage measurements is built into application as a shared library, so it's unusable for end-users. Due to the way `make/qmake` works, enabling/disabling coverage requires `make distclean` before changing flavour(ie. from release to coverage) so that all object files are rebuild(alternatively one can run `make clean` and `qmake -r`, which will update `Makefiles`).
There are 4 frontends for displaying code coverage added as make targets. Those are: `gcov, lcov, codecov and coveralls`. First two create code-coverage report locally, either as text-files(`gcov`, files are located in `src/` with `.gcov` extension) or html report(`lcov`, located in `src/lcov`). The other two uploads code coverage to external services(respectively [codecov](https://codecov.io/) and [coveralls](https://coveralls.io/)). Please note that `codecov` does not include files with no coverage(due to the bug in gcov), so the total coverage percentage is quite misleading.
In order to be able to see current code coverage:
1. Start with clean repo
`$ make distclean`
2. Reconfigure qmake to create coverage build
`$ qmake CONFIG+=coverage qtpro.pro`
3. Build, run tests and create coverage report
`$ make lcov`
4. View report :)

Also from now, all tests can be run with `make check`.

Travis creates code-coverage report using linux build, as I don't know how to install required stuff(lcov, coveralls-lcov(this is ruby gem)) on osx. Build steps shall be the same.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,7 +12,7 @@ build_script:
   - cmd: cd tests
   - qmake
   - mingw32-make
-  - cmd: auto\util\release\tst_util.exe
+  - mingw32-make check
 artifacts:
   - path: Output\qtpass-*.exe
   - path: src\release\qtpass.exe
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,11 +5,11 @@ install:
 build_script:
   - qmake qtpass.pro CONFIG+=static
   - mingw32-make
+  - mingw32-make check
 #after_build:
   - cmd: copy README.md src\release\README.txt
   - cmd: copy LICENSE src\release\LICENSE.txt
   - iscc qtpass.iss
-  - mingw32-make check
 artifacts:
   - path: Output\qtpass-*.exe
   - path: src\release\qtpass.exe
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,9 +9,6 @@ build_script:
   - cmd: copy README.md src\release\README.txt
   - cmd: copy LICENSE src\release\LICENSE.txt
   - iscc qtpass.iss
-  - cmd: cd tests
-  - qmake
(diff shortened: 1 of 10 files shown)


tezeb:
I closed it not to spam you with too many tries, while I am figuring out what's wrong with Appveyor.
(I probably should have comment what am I doing).

annejan:
Do we need coverage on appveyor, isn't travis enough, coverage differences between platforms should be negligible.
Unit testing there should be on but double coverage reports are for a later stage imo . .

tezeb:
Yeah, but the problem is with build not with coverage.
It fails while building tests, and imho it would be good to run them for all versions.

tezeb:
Almost there. I hit another 2 issues(my lucky weekend I guess).
The first one is that apparently windows build requires resource file(like `windows.rc` for `QtPass`) for each `app` build. 
The other is unfortunately a [bug](https://bugreports.qt.io/browse/QTBUG-6453)(and "unsolvable" one), but there is a workaround!!! Yay!

tezeb:
Ups. I sqashed appveyor test commits and I can't reopen now.
Will create new pull request.

PR #321, opened 2017-09-08
Ignore cmdline arguments if -session is used.

### LINKED ISSUE #320: "-session XXX" upon session restore taken as search string
X window managers pass a session id to programs that they restore. qtpass does not recognize the special cmdline option ``-session SESSION_ID`` and puts the string into the search box (see screenshot).

I'm not calling for session handling, I'd simply like to have this cmdline argument ignored.

qtpass-issue

Achimh3011:
Fixes #320

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -43,7 +43,7 @@ int main(int argc, char *argv[]) {
     text += argv[i];
   }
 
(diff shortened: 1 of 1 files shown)


Achimh3011:
Sorry, referenced wrong issue in commit.

PR #322, opened 2017-10-11
Add sftp, ftps, webdav and webdavs as supported links

cgonzalez:
Tested in KDE Plasma with Dolphin.

(Perhaps it could be better if QtPass allowed to define the supported link types.)


annejan:
The type definition is something for a future feature.

I was looking into changing the RegularExpressions into VerbalExpressions too . .

PR #323, opened 2017-10-14
galego actualizado

xmgz:
fixed galician translation


annejan:
Thanks for the effort @xmgz 👍

PR #324, opened 2017-10-23
Extract static library and separate main function

tezeb:
This commits separates main and rest of the code. Main is then used to created QtPass, using rest of the code compiled into static library. The same static library can now be directly linked with unit-tests, so there is no longer need to add separate object files for each test.
Common options moved from `*.pro` to `qtpass.pri`.


annejan:
Wow, thanks!
This is a much needed cleanup, will look into the Windows build issue tomorrow.

annejan:
Fixed last mac issue, merging . . 
Will look into windows issue(s) when I get to office and can spin up a Windows VM

tezeb:
Thanks for fixing it. I hoped to at least fix the Windows issue, but get carried away. Will set up env and try to fix it.

PR #346, opened 2018-01-09
Fix nested template argument list compile error

martinburchell:
Fixes this compilation error:

In file included from ../../../src/imitatepass.h:5:0,
                 from ../../../src/qtpasssettings.h:5,
                 from ../../../src/storemodel.h:10,
                 from ../../../src/util.h:4,
                 from tst_util.cpp:1:
../../../src/simpletransaction.h:11:54: error: ‘>>’ should be ‘> >’ within a nested template argument list
   std::queue<std::pair<Enums::PROCESS, Enums::PROCESS>> transactionQueue;

### DIFF
--- a/src/simpletransaction.h
+++ b/src/simpletransaction.h
@@ -8,7 +8,7 @@ class simpleTransaction {
 
   int transactionDepth;
   Enums::PROCESS lastInTransaction;
-  std::queue<std::pair<Enums::PROCESS, Enums::PROCESS>> transactionQueue;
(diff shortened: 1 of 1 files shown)


tezeb:
This is not necessary since C++11, which afair is how QtPass shall be
compiled.

Which compiler(version) are you using?

martinburchell:
@tezeb g++ (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

5bentz:
The last commit 2c7fcca reverted this fix ! Causing compilation errors on Ubuntu 16.04 LTS.

PR #347, opened 2018-01-11
New scroll bar on large files

destanyol:
Changed the style of the scroll bar when the text file containing the password is larger than usual.
I have never used Qt and I am on Linux KDE so I don't know if the change is global.

newscroll oldscroll

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -383,7 +383,35 @@
            <bool>false</bool>
           </property>
           <property name="styleSheet">
-           <string notr="true">background: transparent</string>
+           <string notr="true">
+             background: transparent
+             setStyleSheet(QString::fromUtf8(QScrollBar:vertical {              
+              border: 1px solid #999999;
+              background:white;
+              width:10px;    
+              margin: 0px 0px 0px 0px;
+             }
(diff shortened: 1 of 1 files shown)


annejan:
Works on macOS as well..

screen shot 2018-01-12 at 02 00 12
Before

screen shot 2018-01-12 at 02 01 00
After

lukedirtwalker:
This change changed the background of file content to white. Was that intended? I liked it more how it was previously.

destanyol:
Ouch it's true, on my Linux it didn't change. I will take a look when I have some spare time. 
I'm sorry...

jounathaen:
I'm really not an expert on this, but my first guess would be, that you only have to remove these lines from the commit:
```
+             setStyleSheet(QString::fromUtf8(QScrollBar:vertical {              
+              border: 1px solid #999999;
+              background:white;
+              width:10px;    
+              margin: 0px 0px 0px 0px;
+             }
```

destanyol:
Sorry I forgot about this...
Today I have tried to solve the changes on the background color but I have never used QT before and I have failed. I'm so sorry...
I have created a new pull request reverting the changes done here.

PR #349, opened 2018-01-19
Fix typo in french translation

babolivier:
Fixed a common misuse of the french word "décrypter" in the french translation.


kenji21:
👍

PR #359, opened 2018-03-12
Fix & make clearClipboard more robust

lukedirtwalker:
The current version had 2 bugs:
1) The clippedText was never cleared, that isn't that much of a problem
since it is overriden anyways
2) The condition to check if clearing should happen was always true
since it compared the clippedText member variable with itself. The
clippedText variables that are create above the condition are out of
scope.

I fixed these bugs and clear both clipboard types if they contain the
given text, ignoring the setting, which should make it more robust.

### DIFF
--- a/localization/localization_ar_MA.ts
+++ b/localization/localization_ar_MA.ts
@@ -516,14 +516,14 @@ Expire-Date: 0
     <message>
         <location filename="../src/mainwindow.ui" line="72"/>
         <location filename="../src/mainwindow.ui" line="75"/>
-        <location filename="../src/mainwindow.cpp" line="1333"/>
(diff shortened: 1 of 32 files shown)


lukedirtwalker:
I'll rebase to fix the conflicts

lukedirtwalker:
I have some more changes I will want to commit. Should I just create a new PR? Or append here?

annejan:
Appending here is fine by me @lukedirtwalker

lukedirtwalker:
@annejan added some more commits, they don't really have to do with the initial fix. I think it is best to review commit by commit.

annejan:
I will later tonight . .

PR #360, opened 2018-03-15
Refactoring

lukedirtwalker:
Introduce the FileContent class for parsing the password file and use it where we parsed manually before. Also slightly change the behavior of MainWindow how it shows the content (see commit for more details).


annejan:
I'm sorry I didn't have time earlier . . 

Hope you have not planted a red-herring I might have missed, but it all looks good to me 👯‍♂️

Issue #300, opened 2017-03-02
Removal of files outside of password-store


annejan:
Received via mail:

> One day I was trying out password managers, I came across QtPass. It seems nice but once when I wanted to delete a file, it showed me "Delete /?", I assumed that it just meant to remove all the folders within the directory, that is, delete all my passwords. So I chose it, it deleted every file off my computer, didn't put it in trash, just gone.

Looking up the sender told me this was most probably on Ubuntu linux.
Have tried to reproduce, but couldn't so-far.

annejan:
The affected user sent me his QtPass.conf file, that only contained [mainwindow] layout related information.

It seems that QtPass was never properly initialised.
But that should not matter. . QtPass should never be able to manipulate files outside of the (current) password-store folder.

I'm keeping this ticket as a reminder to write some safeguards on the (native) handlers to never be able to operate outside of the store.

jounathaen:
Hmm, just wondering how/if it could remove files outside the home folder, as when running the program as non-su you wouldn't have permissions to do so. My guess would be, that the home folder was choosen as password-store. (Maybe qtpass should catch initializing the home folder as pwd-store with a pop up warning?)
But I'm just guessing.

FiloSpaTeam:
I think QtPass should check files to remove before cleanup. So even if you have photo or videos in this folder the delete method will skip those files.

Which types of file QtPass handles? Only gpg files?

annejan:
There are two file types . .  `.gpg` and `.gpg-id`

FiloSpaTeam:
Ok i found a way. The problem itself is "pass" because the command rm doesn't accept specific arguments like Test/*.gpg so i need to:
* select all keys under that folder and delete singularly
* check if dir is empty and delete them if needed

It that case no external pass file are deleted accidentally and if you mix something you are safe.

annejan:
My main question is, why would there ever be non pass related files in your `.password-store` folder?

FiloSpaTeam:
I have the same question LoL but i think that someone doesn't trouble if something goes into this folder (it's a folder anyway).
I imagine that: if i created my wallet in the home_path (i understand is not a good thing) but i don't know very well how it works or something else and i need to delete my 2 test passwords, i lose everything and it's unclear because in the UI those files (my files) are filtered out.

I think it's a related to bad user experience more than i should do or i shouldn't do this.

annejan:
[...]

PR #363, opened 2018-04-04
fix for #300

FiloSpaTeam:
Now before folder deletion, check if there are unexpected files and warn the user.


FiloSpaTeam:
I added a new QAction to open folder directly from QtPass using QDesktopServices

PR #376, opened 2018-04-28
Some fixes and refactoring.

FiloSpaTeam:
I tryied a big refactoring but first we need to clean up classes from duplicated code or similar code.


annejan:
I really love all the work you are doing here!

PR #392, opened 2018-05-11
Fix High Dpi Support. Works now under Windows and KDE/Plasma.

hgraeber:
Commit 4a530820 was incomplete...

Qts scaling environment variables shall never be changed by application.
That way the user hasn't any chance to fix things himself. With setting
QT_AUTO_SCREEN_SCALE_FACTOR to 1 QtPass does not work under KDE/Plasma
and cannot even be fixed by setting Qts scaling environment variables.

### DIFF
--- a/main/main.cpp
+++ b/main/main.cpp
@@ -39,7 +39,6 @@
  * @return
  */
 int main(int argc, char *argv[]) {
-  qputenv("QT_AUTO_SCREEN_SCALE_FACTOR", "1");
   QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
   QString text = "";
   for (int i = 1; i < argc; ++i) {


hgraeber:
@FiloSpaTeam Why. The only reason can be that Qt 5.6s HighDPI is broken.

By forcing the QT_AUTO_SCALE_FACTOR environment variable to a fixed value you will break it even more, because it is not possible anymore to use the variable outside of QtPass to fix the behavior on some platform where HighDPI misbehaves.

So I can only repeat: It is better to **never** change those environment variables inside of the program.

hgraeber:
@FiloSpaTeam With Qt before 5.6 HighDPI is broken anyways. For distributions using those old version I would expect that they patch it themself. I think they already have many other patches of this type inside of their packages and HighDPI still doesn't work completely right anyways.

PR #407, opened 2018-07-23
Continuing refactoring

FiloSpaTeam:
What i'm trying to do:
* MainWindow emit signal and connects various Dialog
* QtPass class connect signal to Pass and various QtPassSettings

What i need to do:
* Move mountWebDav inside QtPass class
* Try to move outside checkConfig (or before MainWindow or later render)


annejan:
Oh wow, those are nice changes . .

FiloSpaTeam:
Before merge i want to test more and try to do what i wrote above :+1:

FiloSpaTeam:
What do you think? :)

lukedirtwalker:
If I may: I would like it if you would squash the commits. It doesn't look nice to have a lot of "Saving" commits in there.

FiloSpaTeam:
Oh yes :) i ll do it asap

FiloSpaTeam:
Travis CI build failed: possible this? -> -platform name

> This command line argument applies to all Qt applications, but might be especially useful in the context of auto-testing. By using the "offscreen" platform plugin (-platform offscreen) it's possible to have tests that use QWidget or QWindow run without showing anything on the screen. Currently the offscreen platform plugin is only fully supported on X11.

Taken from https://doc.qt.io/qt-5/qtest-overview.html

FiloSpaTeam:
I've done a little rebase but for the other commits: it's a lot of work and i don't know finally if it can cause some troubles for the others and for the repo itself :~~
What do you think?

annejan:
I think I'll do some testing on some VMs tomorrow . . 
Since I'm mostly doing data normalisation tasks I think I have time to do a lot of physical testing as-well.

annejan:
I feel confident merging these changes . .

The only interesting thing I noticed is that the sorting order in the tree view is now z-a 😺

FiloSpaTeam:
LOL i didnt noticed. I ll see and fix anyway :) if you feel i need to change something tell

FiloSpaTeam:
Before merge i need to solve:
* QTreeView sorting (however is Ascending...)
* Solve a BUG when you type fast into lineEdit, sometimes root path is visible and random files are taken (this is present in the v1.2.3 too) xD

FiloSpaTeam:
To solve the second one, I created a Timer of 350ms that start on lineEdit changed and fire on timeout the real search event.

wamtek:
@FiloSpaTeam It seems that something with the copy function is not working as expected in your version. After i decrypted an item from the list the password is automatically copied to the clipboard. But when i click on the copy symbol next to the password nothing happens. No matter if i try to copy the password or any other entry.
The unmodified version copies fine when clicking the symbol.

FiloSpaTeam:
Ok perfect, i'll fix it :) i didn t tried the copy button lol

wamtek:
Copy button works again. :+1:

PR #409, opened 2018-08-05
pwgen: fix inverted "Generate ... less secure passwords" checkbox

ahippo:
Checking the checkbox "Generate easy to memorize but less secure passwords"
(aka `isLessRandom()`)
means _omitting_ "--secure" option to `pwgen`.
Unchecking the checkbox means _passing_ "--secure" option to `pwgen`,
which is meant to be the default for QtPass.

This got broken in
b11e71d331310c27fde3e53a098ba108bc8f0f61 Use new executor in Pass

Fix this by inverting the check once again.


ahippo:
Thank you for merging this so quickly!

PR #413, opened 2018-08-25
Bugfixes

rdoeffinger:


src/mainwindow.cpp
```diff
@@ -411,6 +411,9 @@ void MainWindow::config() {
       this->show();
 
       updateProfileBox();
+      // For freshStart, proxyModel is not yet configured
+      // and maniplating it will assert
+      if (!freshStart)
       ui->treeView->setRootIndex(proxyModel.mapFromSource(
```
lukedirtwalker: Please indent the line after the if. Otherwise this looks very confusing.

PR #416, opened 2018-08-30
Support for WSL on Windows - NOT FOR MERGE as-is

rdoeffinger:
This is likely still a bit messy and it includes the patches from the previous merge, so not quite suitable to just merge.
But I implemented support for using WSL to use Linux binaries of gpg, git, pass etc. on Windows.
This includes auto-detecting and using those if no others are available.
Despite the auto-detection it might be a good idea to document it though...

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -538,7 +538,7 @@ void ConfigDialog::wizard() {
 
   QString gpg = ui->gpgPath->text();
   // QString gpg = mainWindow->getGpgExecutable();
-  if (!QFile(gpg).exists()) {
+  if (!gpg.startsWith("wsl ") && !QFile(gpg).exists()) {
     criticalMessage(
         tr("GnuPG not found"),
         tr("Please install GnuPG on your system.<br>Install "
@@ -555,7 +555,7 @@ void ConfigDialog::wizard() {
   dbg() << names;
 #endif
 
-  if (QFile(gpg).exists() && names.empty()) {
+  if ((gpg.startsWith("wsl ") || QFile(gpg).exists()) && names.empty()) {
     KeygenDialog d(this);
     if (!d.exec())
       return;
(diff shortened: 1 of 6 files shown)


rdoeffinger:
I intend to clean this patch series up and submit separately, but here is what I was working on:
https://www.microsoft.com/store/apps/9N8GQNMKBWX7
It is only reachable via that link, not through search, so there's no problem removing it if you don't like the idea. I'd be happy to hear any thoughts on it though.

jounathaen:
I don't like Microsoft and I don't like the Windows Store. But if it works and is easy to maintain, why not :+1:

Some questions: Who would maintain that build? And who is legally responsible for the store package? And does Microsoft allow distribution of GPL3 Programs and is the license compatible with the store?

rdoeffinger:
> Who would maintain that build?

That's one of the questions I guess. It is not significantly different from the normal release, so maybe it would make sense to handle it together with the normal Windows releases.

> And who is legally responsible for the store package?

If anyone really wants to know they'd have to ask a lawyer. Primarily Microsoft I guess, though they will do their best to deflect any issues onto someone who agree to the store agreements.

> And does Microsoft allow distribution of GPL3 Programs

Have not seen anything that would forbid it so far at least.

> and is the license compatible with the store?

No issues I am aware of. Windows S is possible to convert into a full version (and not supported by the app anyway), and on all other Windows versions it is possible to install self-signed apps, it is possible to install even unsigned apps (if unpacking them first), and it is possible to run the program just as program without involving any "app" stuff at all.
This is not a UWP version or anything, it's just a different way of distributing a completely ordinary desktop app.
I should however put actual links to source code in the app description before putting it live if we go that way.
Or is there any other issue you are thinking of?

rdoeffinger:
Sorry, this one got a bit messy, I will support pull requests issue by issue shortly.

PR #418, opened 2018-09-13
Dutch (nl) translation improvements

equaeghe:


equaeghe:
Hmm, I just realized that also the `qtpass.pro` and `qtpass.qrc` files probably also need to be updated. I'll hold off from doing that until I get some feedback, as there may be other things I've missed.

annejan:
I think those are the only needed changes, although I will have to see if the language loader works like this since I think it's based on language_COUNTRY (like the nl_NL, nl_BE etc).
Will probably have time to play with this tomorrow.

equaeghe:
> I think those are the only needed changes,

I searched through the source files for nl_NL and only the desktop file also needed some attention.

> although I will have to see if the language loader works like this since I think it's based on language_COUNTRY (like the nl_NL, nl_BE etc).

Hmm, I hope it works (should've tested it myself, sorry); for ‘ca’ (Catalan?) there is also no country code given currently.

> Will probably have time to play with this tomorrow.

Ok, let me know if I need to fix something.

PR #419, opened 2018-09-25
Tested working on Mac Os HS with pinentry-mac

riccardocossu:
No need to install gpg tools in Mac Os, pass and pinentry-mac will do the job just fine


annejan:
Tested and works for me too 💃

PR #421, opened 2018-09-29
Display passwords as QR codes

frawi:
While there are apps to use pass on an Android phone, I don't feel comfortable with that. It requires GPG on my phone, and I want to use only a small subset of the passwords on my phone. This is all possible, but I think I don't need a complex solution. Most passwords will be stored in the corresponding app anyway.

So I was often using qrencode to copy passwords to my phone. This allows me to control which passwords are on my phone very easily, and allows me to use complicated passwords. I mostly using something like the following command to get passwords out of pass to the screen:

`pass wifi/home | qrencode -o- | display`

Since I mostly use QtPass as GUI, I thought it would be nice to have this feature built in.

This pull request is my implementation of this feature. I am not very familiar with Qt, I tried my best to keep the changes minimal but complete. I made the following modifications:

* added an icon for the QR code
* the icon will be displayed next to the "copy to clipboard" icon
* when the icon is clicked, a new window is opened containing a PNG with the QR code of the text
* the QR code is generated by calling the external tool qrencode https://github.com/fukuchi/libqrencode/
* text input and png output are passed through stdin/stdout, so there should be no leaks on the file system
* the feature is configurable in the config dialog, similar to OTP
* if qrencode is not installed, the feature will be disabled and a tooltip is displayed in the config

I used it for some time now and thought it works quite well. There are some issues where I am currently not so sure how to solve them:

* Opening a new window is nice, because it allows resizing, but maybe it is not the best way to present the code. Maybe the window should also close after a certain timeout.
* The password is passed to the external program qrencode. If this program is malicious, it could sniff the passwords.
* The QR code could be generated directly in QtPass, but this would probably add more dependencies.

Please give me your feedback. I would be very happy if this is adopted.

### DIFF
--- a/icons/qrcode.svg
+++ b/icons/qrcode.svg
@@ -1,0 +1,133 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="24"
+   height="24"
+   viewBox="0 0 24 24"
+   version="1.1"
+   id="svg6"
+   sodipodi:docname="qrcode.svg"
+   inkscape:version="0.92.3 (2405546, 2018-03-11)">
+  <metadata
+     id="metadata12">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs10" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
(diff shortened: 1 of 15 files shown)


jounathaen:
I basically like the idea.
I have not tested it or looked at it in detail, but a few comments from my side:
- This would make libqrencode a (loose) dependency of QtPass. Are there any other tools for generating qrcodes or is this the standard?
- How common is libqrencode? Can it be installed with common packet managers?
- You hardcoded the feature for non-apple/non-windows installations. Why? Can't libqrencode be installed on those Systems or didn't you test it on these?

My comments on your questions:
>  Opening a new window is nice, because it allows resizing, but maybe it is not the best way to present the code. Maybe the window should also close after a certain timeout.

My main concern would be, that the QR code should only open when a button is pressed. If it opens by default, it would make the password more vulnerable to "over-the-shoulder" viewing cameras. (I guess you already implemented it in this way.)

> The password is passed to the external program qrencode. If this program is malicious, it could sniff the passwords.

Yes, but if you have malicious software on your computer, everything is lost anyway. It could act as a keylogger and steal your keychain password, it could make screenshots of QtPass and read the password,...

> The QR code could be generated directly in QtPass, but this would probably add more dependencies.

I think so too. QtPass already uses external programs like pass or pwgen. Another optional external dependency does not hurt.
On the other side, there is some effort in faking these external programs for Windows, as they can't be used on that platform...

annejan:
> You hardcoded the feature for non-apple/non-windows installations. Why? Can't libqrencode be installed on those Systems or didn't you test it on these?

On macOS it's as simple as `brew install qrencode` unfortunately `display` does not work out of the box on Apples.

thegrandpoobah:
is this supposed to work on Mac OS? I love it on Linux, but can't seem to enable it on Mac?

annejan:
Only in upcoming release will it work on mac..

thegrandpoobah:
ahh.. okay thanks! looking forward to it :)

PR #424, opened 2018-10-24
Update to prevent the installer requesting admin

hughwilliams94:
Added PrivilegesRequired=lowest as per http://www.jrsoftware.org/ishelp/index.php?topic=setup_privilegesrequired


annejan:
Thank you very much @hughwilliams94

ahippo:
This seems to also break updates.
image

Issue #463, opened 2019-08-12
Renaming password


annejan:
Currently not in the gui

Issue #446, opened 2019-02-07
[Feature Request] Edit main title field


Spheerys:
Hi,

Is there a reason why we can't change the entry name/title (finally, the filename) from qtpass ?
If not, could we imagine to do it in future version ?

Thanks :)

jounathaen:
Yes, this should be implemented.
BTW: I think this would be an easy "first PR" if somebody want's to.

annejan:
Related #463

PR #469, opened 2019-08-29
Add support for passwords and directories renaming as requested in #463

Noettore:
Added support for renaming passwords and directories as requested in #463 and #446 
Right-clicking an entry shows a new "rename" action in the context menu.
The action opens a dialog which lets you modify the actual entry name.
The renaming is performed via the Pass::Move() method.


annejan:
Thank you, this is very useful!

Issue #404, opened 2018-07-18
Profile Selection Drop Down does not sort - random order


jkman340:
In our use of QTPass we have multiple profiles configured for various stores.  The drop down button to select the active profile (top right of main window) does not sort the profiles and appears to be random when QTPass is launched.   Requesting to have the drop down list sorted alphabetically to make it easier to find a profile when multiple profiles. are configured. 
We are using QTPass on Ubuntu 16.04 and 18.04.

PR #472, opened 2019-09-01
Sorted profiles dropdown as in #404

Noettore:
Now the profileBox is alphabetically sorted as requested in #404


annejan:
Thank you, this is very helpful :)

Issue #327, opened 2017-10-27
Add otp (two factor authentication) support


figgyc:
This is just a minor feature request.

annejan:
@figgyc OTP secrets can be stored in `pass` or QtPass

If you mean OTP for unlocking `gpg` I'm afraid the request is out of scope, since unlocking `gpg` is handled by the `gpg` subsystem.

figgyc:
I know that what i'm trying to say is is there a way for qtpass to generate secrets from the code?

annejan:
There seems to be a `pass` extension for OTP handling https://github.com/tadfisher/pass-otp

jsmestad:
Would love to see this happen!

PR #499, opened 2019-12-20
Don't base pass-otp availability decision on hardcoded /usr/lib

nh2:
`/usr/lib` does not exist in various Linux distributions, nor does it work if you have installed your qtpass to a custom prefix such as `/usr/local` or into your home directory.

Disable this check for now, so that the OTP functionality can be enabled in such cases.

As a result, `ConfigDialog::isPassOtpAvailable()` now only has the functionality to return `false` on platforms where the OTP functionality is certainly not supported.

CC @FiloSpaTeam as the author of https://github.com/IJHack/QtPass/commit/5fc6c652af6b6cc949b571c090d5e6e65228b135 (via #407); also mentioning #394 and #327 so that this PR shows up there for others that, like me, searched for this problem.

---

A better future detection logic might be to run `pass otp --version` and check if that succeeds, but I'm not doing that here because it's more sophisticated and I'd like qtpass to work well with `pass-otp` for all users as soon as possible.

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -567,9 +567,7 @@ bool ConfigDialog::isPassOtpAvailable() {
 #elif defined(__APPLE__)
   return false;
 #else
-  QFileInfo file("/usr/lib/password-store/extensions/otp.bash");
-
-  return file.exists();
+  return true;
 #endif
 }
 


maciejsszmigiero:
I think the pass OTP extension path should simply be made (compile-time) configurable,
so distributions can set what path they install this extension at.
OTOH, if you compile QtPass yourself you can set it as you like, even to /bin/true.

OTP module autodetection is useful, please don't remove this functionality.

nh2:
> I think the pass OTP extension path should simply be made (compile-time) configurable,
> so distributions can set what path they install this extension at.

That'd do as well.

I didn't do it here because it's more sophisticated and I won't have time to do it.

While I'd be happy with that, I also am not fully convinced that that's how `pass` envisioned it to be used; from `man pass` you can set `PASSWORD_STORE_EXTENSIONS_DIR` to anything you like at run time, thus also allowing user-provided extensions (for example in `~/.pass/.extensions`) that are not available at compile time. That's why I think that checking e.g. `pass-otp --version` may be even better than your proposed `/bin/true` solution.

> OTP module autodetection is useful, please don't remove this functionality.

Personally I think it's a good approach to make a quick fix available immediately and work on an improvement afterwards (especially for a convenience feature like force-_disabling_ a check box), versus having it broken until the improved solution is available.

We must remember that the only thing this check does is to prevent the user to turn on an off-by-default checkbox by hand (or at least it seemed to be off-by-default on my system), which I'd consider not really that useful compared to not being able to use it at all.

annejan:
> Personally I think it's a good approach to make a quick fix available immediately and work on an improvement afterwards (especially for a convenience feature like force-disabling a check box), versus having it broken until the improved solution is available.

I agree with this . .

PR #526, opened 2020-05-23
Fix blurry icons when fractional scaling is enabled

mthw0:
...by adding UseHighDpiPixmaps flag to the main.cpp file. Currently icons are blurry with fractional scaling enabled. This change fixes this issue. I built and tested this on KDE Plasma with fractional scaling at 1.25.


annejan:
Thank you, awesome!

Issue #514, opened 2020-03-06
Display passwords in mono space font


InnayTool:
Sometimes is hard to decide if a character in a password is for example an upper case i or an lower case L. So it would be cool if this could be displayed in a mono space font.

neilmayhew:
I see that the changes for this were [reverted][1] recently because they stopped working. However, the problem that this issue identifies is back again. Are there plans to do anything about it?

[1]: https://github.com/IJHack/QtPass/commit/0595f17a71882e8884bfc638642873a44049be66#diff-c72b037f3ac7abbd3d0ee519b6f456013a9bbf5a59df16308939320c04970447

neilmayhew:
It looks to me like `font-family: monospace` could be added to the `setStyleSheet` call at [mainwindow.cpp:1107](https://github.com/IJHack/QtPass/blob/4f4e0c03357acc563f691a82ac4bbecdb08e4eab/src/mainwindow.cpp#L1107).

PR #527, opened 2020-06-12
Update minimum Qt version

cmol:
The code uses QRandomGenerator which was not introduced until Qt 5.10


cmol:
I accidentally pushed again to this branch for a fix to #514.

annejan:
Thank you for this update, I completely missed this dependency.

PR #528, opened 2020-06-12
#514 Show password with a monospace font

cmol:


annejan:
This might save some lI| and other mistakes in reading.

PR #529, opened 2020-06-12
Clear search on profile change

### LINKED ISSUE #524: [Request] clear search password when change profile
Hi,
we use QtPass and we also use multiple profile. When I search for a password and later change the profile the search term is still there and the list is empty because there is no password.. 

May you can clear the search text box when the profile is changed?

Thanks and best regards!

cmol:
Fixes #524


annejan:
This works like a charm and is indeed a well needed fix for #524

PR #532, opened 2020-06-26
Fix issues with renaming passwords and moving folders

### LINKED ISSUE #487: Renaming passwords and directories fail
On renaming of a directory:
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
Git status shows an uncommited but staged change for the folder

On renaming of a file:
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
Git status shows an uncommited but staged change for the file. The file is unencrypted!

ChaoticEnigma:
Fix #487
Rewrite ImitatePass Move method
If no file is provided to ImitatePass GitCommit, do not pass --


ChaoticEnigma:
Placeholder PR - I'll update here if I need to fix anything

PR #547, opened 2020-10-18
Move MainWindow to the screen the cursor is on

JFK422:
A small quality of life improvement, that moves the MainWindow, upon launch, to the screen the mouse cursor is on.
This is useful on multiple monitors, so the MainWindow doesn't appear on a "random" screen.


annejan:
Thank you!

Issue #520, opened 2020-04-20
UI can't handle passwords with periods in their name


gardenappl:
Let's say, I have a password named `my.account`, so it's stored in a file named `my.account.gpg`. I want to rename it to `different-name`, so I click on "Rename password".

- In the "Rename file" window, instead of saying "my.account" it just says "my".
- When I enter the name "different-name", instead of saving it as `different-name.gpg`, it saves it as `different-name.account`! Then it no longer shows up on the password list.

Tested on Windows 10, without Git or native Pass.

woodenphone:
I have also encountered this but while adding website passords.
"foobar.com" creates a file named "foobar.com" instead of a file named "foobar.com.gpg"
Manually changing the file extention of the password file(via windows explorer) to end in .gpg makes the password entry show up in the UI again.

Encountered in Windows 10, QTPass version 1.3.2, manually configured gpg "C:/Program Files (x86)/GnuPG/bin/gpg.exe" git "C:/Program Files/Git/bin/git.exe" pwgen "C:/Program Files (x86)/PWGen/PWGen.exe", native git/gpg option used, password store path "C:/Users/USERNAME/password-store/", "Use git" checked, "Automatically add .gpg-id" checked", "Automatically push/pull" both unchecked.

unor:
On GNU/Linux (using Pass, but not using Git):

* Adding "test2.foo" creates the file `test2.foo.gpg`  
 Adding "test3.foo.bar" creates the file `test3.foo.bar.gpg`

* When trying to rename "test2.foo", the UI dialog only shows "test2"  
 When trying to rename "test3.foo.bar", the UI dialog only shows "test3"

* When renaming the entry "test2.foo" (which shows as "test2" in the renaming UI) to "bird", the entry gets renamed to "bird.foo", and the file gets renamed to `bird.foo.gpg`
 When renaming the entry "test3.foo.bar" (which shows as "test3" in the renaming UI) to "fish", the entry gets renamed to "fish.foo.bar", and the file gets renamed to `fish.foo.bar.gpg`

So, on GNU/Linux the issue of the renaming UI is the same as described above for Windows, but the files get renamed correctly, so the entries do not disappear in the list UI.

ichthyosaurus:
> When I enter the name "different-name", instead of saving it as `different-name.gpg`, it saves it as `different-name.account`! Then it no longer shows up on the password list.

... and

> I have also encountered this but while adding website passords.
[...]

PR #558, opened 2020-12-20
Explicitly only remove ".gpg" when renaming files

ichthyosaurus:
QFileInfo::baseName removes any path plus the complete suffix. The alternative QFileInfo::completeBaseName would only remove the last suffix (e.g. ".gz" in "foo.tar.gz") but we want to remove ".gpg" only.

Cf. https://doc.qt.io/qt-5/qfileinfo.html#completeBaseName

This fixes part of #520. I could not reproduce the other errors mentioned there ("different-name.gpg" → "different-name.account" and "`foobar.com` creates a file named `foobar.com` instead of a file named `foobar.com.gpg`").

(Only tested on GNU/Linux.)

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1018,7 +1018,10 @@ void MainWindow::renamePassword() {
   bool ok;
   QString file = getFile(ui->treeView->currentIndex(), false);
   QString filePath = QFileInfo(file).path();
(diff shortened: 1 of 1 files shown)


annejan:
Thank you very much for your contribution!

Tested on MacOS and Windows (VM)

Issue #586, opened 2022-02-13
The password display should not distinguish the first line from all following


l3u:
Currently, the password display consists of a "Password" label followed by the first line of the decrypted password file and a text area containing all following lines.

The pass program itself does not distinguish between single-line and multi-line password files. Every entry is a file, no matter how many lines are stored.

This leads to unexpected behavior, at least for me as a console pass user. First thing is that the first line of a pass file is not necessarily a "Password", and the second thing is that – for multi line files – I format my stuff to be lined up indended by spaces. This problem could easily be solved by using a fixed-width font, as already requested in issue #577. I personally would also make that font configurable to that a "good" one can be chosen, with definitely distinct characters.

But apart from the font issue, the decrypted file should be displayed as-is: Line by line, the first line in the exact same way all following lines are displayed. Just as pass does it on the console.

E.g. if you store a 63 character long WPA3 key, it's squitched inside that one-line line input, clipped on the top and bottom, along with an up- and down-arrow. Or if you save some file with multiple lines, and the first line is just a header, it is displayed as the "Password", and the rest follows below in another widget.

I'd probably simply use a QTextEdit with a monospace font and no line wrapping …

l3u:
Everthing bitched about here is fixed/added in PR #587 ;-)

annejan:
The first line (password) in the display panel is styled differently to distinguish it from other content. This is intentional design. If you have specific styling concerns, please provide more details about what you'd like to see changed.

Issue #577, opened 2021-12-17
Passwords are displayed in a font that makes it hard to identify certain characters


Kritzefitz:
**Is your feature request related to a problem? Please describe.**

Passwords are displayed in a proportional sans serif font both in the display field as well as in the edit input field. This makes it really hard to identify certain characters such as lowercase L or uppercase i, as shown by the following screenshots.

![image](https://user-images.githubusercontent.com/8547449/146549099-1b63d53c-d1f2-4d81-b27c-30fc5beca49a.png)

![image](https://user-images.githubusercontent.com/8547449/146549157-bc154783-8f4d-4751-ba9f-4e88cc391b1a.png)

**Describe the solution you'd like**

Passwords should be displayed in a monospace font. This usually makes it possible to identify each individual character. This should also affect the input field for editing passwords.

**Describe alternatives you've considered**

Using a proportional font with clear visual separation for each character would also be an option. But most monospace fonts are consciously designed to do so, so they should be the easiest solution.

As a workaround it is currently possible to paste the password into some other application that uses a monospace font (such as a terminal). But this is of course a very clunky solution.

**Additional context**

This has apparently been implemented in the past in ce4cbde592fc6a57bbacf85e963432b53ab9bd33 but later reverted in 0595f17a71882e8884bfc638642873a44049be66.

ilardm:
worth mention one of the possible solution in #514

PR #587, opened 2022-02-19
Add more options for the password displaying

l3u:
This addresses both issue #586 and issue #577.

I added two additional checkboxes to the configuration dialog:

* "Use a monospace font" that, if activated, sets the system's default monospace font for both the `QTextBrowser` and all `QLineEdit`s used to display password files
* "Display the files content as-is" that, if activated, skips the parsing of the pass file content and simply puts everything into the `QTextBrowser`, as `pass` does when used on the console.

Imo this won't hurt, as the default behavior remains unchanged, but it's convenient for console pass users that want to have the same behavior in QtPass.


l3u:
Just to say it: I adapted the way you did it, like

    void QtPassSettings::setNoLineWrapping(const bool &noLineWrapping) {
      getInstance()->setValue(SettingsConstants::noLineWrapping, noLineWrapping);
    }

but in such a case, it's not really meaningful to use a const reference to a bool value. The pointer to the bool value is bigger than the value itself, so it's actually worse than simply passing the variable by value ;-)

But this is another story …

annejan (approved):

PR #634, opened 2023-06-16
Add pass store signing key feature

### LINKED ISSUE #624: support `PASSWORD_STORE_SIGNING_KEY` with profiles
**Is your feature request related to a problem? Please describe.**

we are evaluating the use of pass/QtPass in a team and are trying to figure out a trust model that would not force all team members to learn how to use git. the encryption with defined OpenPGP keys per unit, like pass/QtPass are offering, is exactly what we would like to use, since we are already using OpenPGP anyway. but IIUC, everyone with access to the password store could add key IDs to any .gpg-id files.

**Describe the solution you'd like**

in the pass docs, the environment variable `PASSWORD_STORE_SIGNING_KEY` is listed with the description:

> If this environment variable is set, then all .gpg-id files and non-system extension files must be signed using a detached signature using the GPG key specified by the full 40 character upper-case fingerprint in this variable. If multiple fingerprints are specified, each separated by a whitespace character, then signatures must match at least one. The init command will keep signatures of .gpg-id files up to date.

QtPass already supports profiles to switch between password stores, which is a neat feature. unfortunately, it currently doesn't seem to supprt setting the `PASSWORD_STORE_SIGNING_KEY` variable. i think this would need to be set independently per profile to be really useful, i.e. add a "select signing keys" button in a new row next to the profle path.

**Describe alternatives you've considered**

it's not really an alternative, but we are discussing limiting the write access to the password store to designated admins to mitigate the potential damage. we would like to combine this with the signing of .gpg-id files.

timegrid:
# Reason
The [pass store signing key](https://man.archlinux.org/man/pass.1.en#PASSWORD_STORE_SIGNING_KEY) feature would enhance the security of qtpass/pass when used by teams. Teams need fine-grained control over the `.gpg-id` user lists of subpaths/stores, but they also have to be able to verify the `.gpg-id` user lists to prevent accidental/malicious privilege escalation. By setting the `PASSWORD_STORE_SIGNING_KEY` envvar to a trusted team signing key id, password admins can provide prepared and signed `.gpg-id[.sig]` files and team members can easily enable the verification by configuring the new signing key id in the corresponding profile. If the verification fails, the error `Signature for [...]/.gpg-id is invalid.` is visible and further actions are prevented. This PR would fix #624, see further discussion there.

# Changes
Adds the [pass store signing key](https://man.archlinux.org/man/pass.1.en#PASSWORD_STORE_SIGNING_KEY) feature by
- changing the **profile type** from `QString` to `QHash<QString, QString>` in order to be able to represent
  more key/value pairs for one profile (`profile['path']`, `profile['signingKey']`)
- changing the **profiles type** from `QHash<QString, QString>` to `QHash<QString, QHash<QString, QString>>`
  in order to store the new data structure
- changing the **profiles settings** from plain keys (childKeys) to a group (childGroups) in order to store more 
  key/values for one profile (`profile/NAME/path`, `profile/NAME/signingKey`)
- adding a column to the **profile ui table**
- adding a **new setting** for `passSigningKey` for the current value of the profile
- updating the **environment** on profile switch in `updateEnv()` (adding/updating the envvar 
  `PASSWORD_STORE_SIGNING_KEY` to the value from the `passSigningKey` setting or removing the envvar,
  if the `passSigningKey` setting is empty)
- adding a **comment** to the `README.md`

# Notes
- ~~Unfortunately the change of profile(s) data structure results in an **empty profile list** on upgrade, so this PR might
  be considered a **breaking change (consider this, if you want to test this PR with your real qtpass configuration)**. But to introduce another parallel data structure would have been awkward/messy
  and this way in the future it is easy to add further key/value pairs for profiles as needed.~~
- ~~This is `pass` only. Using git/gpg would be possible, but quite more time-consuming~~
- After `qmake`, all translation file items included additional location filenames for header files `../src/*.h`. I saw #606,
  but I have no idea, why this happened again and I'm unsure how to proceed, as some line number references
  need to be updated
- All tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ qmake && make && make install
 ## Using profiles
 
 Profiles allow to group passwords. Each profile might use a different git repository and/or different gpg key.
+Each profile also can be associated with a pass store singing key to verify the detached .gpg-id signature (pass only).
 A typical use case is to separate personal and work passwords.
 
 > **Hint**<br>
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ qmake && make && make install
 ## Using profiles
 
 Profiles allow to group passwords. Each profile might use a different git repository and/or different gpg key.
-Each profile also can be associated with a pass store singing key to verify the detached .gpg-id signature (pass only).
+Each profile also can be associated with a pass store singing key to verify the detached .gpg-id signature.
 A typical use case is to separate personal and work passwords.
 
 > **Hint**<br>
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -97,7 +97,7 @@ ConfigDialog::ConfigDialog(MainWindow *parent)
   useTemplate(QtPassSettings::isUseTemplate());
 
   ui->profileTable->verticalHeader()->hide();
-  ui->profileTable->horizontalHeader()->setStretchLastSection(true);
+  ui->profileTable->horizontalHeader()->setSectionResizeMode(1, QHeaderView::Stretch);
   ui->label->setText(ui->label->text() + VERSION);
   ui->comboBoxClipboard->clear();
 
@@ -170,7 +170,7 @@ void ConfigDialog::validate(QTableWidgetItem *item) {
       for (int j = 0; j < ui->profileTable->columnCount(); j++) {
         QTableWidgetItem *_item = ui->profileTable->item(i, j);
 
-        if (_item->text().isEmpty()) {
+        if (_item->text().isEmpty() && j != 2) {
           _item->setBackground(Qt::red);
           status = false;
           break;
@@ -181,7 +181,7 @@ void ConfigDialog::validate(QTableWidgetItem *item) {
         break;
     }
   } else {
-    if (item->text().isEmpty()) {
+    if (item->text().isEmpty() && item->column() != 2) {
       item->setBackground(Qt::red);
       status = false;
     }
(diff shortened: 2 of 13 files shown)


timegrid:
I added a simple migration for the old profile settings, so old profiles are not lost anymore.
*(Caution: If you want to test this, still consider, that the data structure is changed persistently)*

annejan (approved):

timegrid:
I mean those lines, that got added again:
``` diff
M localization/localization_af_ZA.ts
@@ -5,32 +5,39 @@
     <name>ConfigDialog</name>
     <message>
         <location filename="../src/configdialog.ui" line="20"/>
+        <location filename="../src/ui_configdialog.h" line="860"/>
         <source>Configuration</source>
         <translation>Konfigurasie</translation>
     </message>
     [...]
```

timegrid:
Ii also added native gpg/git functions to implement both [signature verification](https://git.zx2c4.com/password-store/tree/src/password-store.sh#n59) and the [signing](https://git.zx2c4.com/password-store/tree/src/password-store.sh#n352) of gpgid file.

thomiel:
We had some testing on Windows and stumbled upon two issues (the first one was kind of inherited and I'd recommend a more general code review to avoind mixing qt-internal and native path separators). This is what I did:

1. Harmoized path separators in `Pass::getGpgIdPath()` as the pass store path couldn't match with the gpgIdDir because of trailing backslashes so two absolute paths were concatenated, leading to checkmarks not set properly in usersdialog.cpp, for example.
2. Added an optional \r in regex of `ImitatePass::verifyGpgIdFile()` to comply with Windows \r\n linebreaks as `QRegularExpression::MultilineOption` won't honor \r as part of a line separator.

https://github.com/timegrid/QtPass/commit/a45da0a326fa7b3541d82bf7f4b93ec3751648e9

PR #790, opened 2026-03-27
feat: use ed25519 for GPG key generation when available

### LINKED ISSUE #406: GnuPG keypair generation takes more than  an hour
The window doesn say whether the process is finished or failed or times out. A status message would be convenient.

nogeenhenk:
## Summary

Fixes #406

### Problem
GPG key generation with RSA can take a very long time (especially with high bit counts) and users may not have feedback on progress.

### Solution
Use ed25519 (ECC) keys by default when GPG 2.1+ is available. ECC keys:
- Are much faster to generate (don't need entropy from /dev/random as much)
- Are more modern and recommended by security experts
- Work with all modern GPG versions (2.1+ since 2014)

Added status messages to show feedback when key generation completes or fails.

### Changes
- Add `Pass::gpgSupportsEd25519()` to check if GPG supports ed25519
- Add `Pass::getDefaultKeyTemplate()` to return appropriate template based on GPG version
- Modify `KeygenDialog` to use dynamic template instead of hardcoded .ui template
- Add status message "GPG key pair generated successfully" on success
- Add status message "GPG key pair generation failed" on failure
- Add "Generate" button to Config dialog for easy key generation
- Add unit tests for ed25519 support check and template

### Templates
- **GPG 2.1+**: Uses EdDSA/Ed25519 for primary key and ECDH/Curve25519 for subkey
- **Older GPG**: Falls back to RSA (same as before)

### DIFF
--- a/localization/localization_af_ZA.ts
+++ b/localization/localization_af_ZA.ts
@@ -129,12 +129,22 @@
         <translation>Sluit spesiale simbole in</translation>
     </message>
     <message>
-        <location filename="../src/configdialog.ui" line="922"/>
+        <location filename="../src/configdialog.ui" line="755"/>
+        <source>Generate</source>
+        <translation type="unfinished">Genereer</translation>
+    </message>
+    <message>
+        <location filename="../src/configdialog.ui" line="758"/>
+        <source>Generate GPG key pair</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <location filename="../src/configdialog.ui" line="938"/>
         <source>Signing Key</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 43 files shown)


src/keygendialog.ui
```diff
@@ -243,21 +243,7 @@
...
-Subkey-Type: RSA
-Name-Real:
-Name-Comment: QtPass
-Name-Email:
-Expire-Date: 0
-%no-protection
-# Do a commit here, so that we can later print &quot;done&quot; :-)
-%commit
-%echo done</string>
+         <string>Template will be set based on GPG version. See Config → Generate GPG key</string>
```
annejan:
```suggestion
         <string>Template contents will be set based on GPG version.</string>
```

PR #1009, opened 2026-04-15
feat: add re-encrypt context menu for directories

### LINKED ISSUE #269: Feature Request: Re-encrypt button
I would like to propose a feature request for a directory level and global level re-encrypt button or that when you give a user access to a folder in qt-pass that it automatically re-encrypts the passwords stored in that directory.

nogeenharrie:
Fixes: #269 

This pull request introduces a new "Re-encrypt" context menu option for directories within the main application's tree view.

Key changes include:
*   **Context Menu Integration:** When a directory is right-clicked, a "Re-encrypt" action is now available in the context menu.
*   **Re-encryption Functionality:** A new `reencryptPath` function handles the re-encryption process.
    *   It first validates the directory path.
    *   It presents a confirmation dialog to the user, explaining that all password files within the selected directory will be re-encrypted using the current GPG recipients defined in `.gpg-id`, and that this action is not easily reversible.
    *   Upon user confirmation, the main window is temporarily disabled to prevent further interaction, and the re-encryption operation is delegated to the `ImitatePass` backend for the specified directory.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1054,6 +1054,13 @@ void MainWindow::showContextMenu(const QPoint &pos) {
     }
     QAction *deleteItem = contextMenu.addAction(tr("Delete"));
     connect(deleteItem, &QAction::triggered, this, &MainWindow::onDelete);
+    if (fileOrFolder.isDir()) {
+      QString dirPath = QDir::cleanPath(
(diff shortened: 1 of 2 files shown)


annejan (commented):
I have tested this locally and it seems to work pretty well 👍🏼

Issue #1026, opened 2026-04-17
chore: clangd static analysis findings


annejan:
## Summary

Baseline clangd static analysis (v21.1.8) run against all source files using `compile_commands.json`.

## Performance issues

Unnecessary value copies — parameters copied for each invocation but only used as `const &`:

- `src/executor.cpp` line 223, 284 — parameter `app`
- `src/mainwindow.cpp` line 1326 — parameter `dir`
- `src/util.cpp` line 120 — parameter `binary`
- `src/gpgkeystate.cpp` line 130 — `record_type` copy-constructed from const ref, should be `const auto&`

## Bug

- `src/qtpass.cpp` line 150 — `-Wunused-lambda-capture`: `this` is captured but not used

## Style / modernize (low priority)

- `modernize-use-trailing-return-type` — lambdas/functions in `mainwindow.cpp` (111, 1062), `pass.cpp` (157, 178, 230), `qtpass.cpp` (150, 158, 549), `usersdialog.cpp` (267, 298)
- `modernize-return-braced-init-list` — `imitatepass.cpp` (513, 798, 818, 826), `pass.cpp` (233, 249, 404)
- `modernize-pass-by-value` + `std::move` — `usersdialog.cpp` line 24
- `modernize-raw-string-literal` — `util.cpp` line 268

## How to reproduce

```bash
./scripts/generate-compile-commands.sh
clangd --check=src/<file>.cpp 2>&1 | grep "^E\[" | grep -v "tweak:\|FAIL:"
```

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1028, opened 2026-04-17
fix: resolve clangd static analysis findings

annejan:
This pull request addresses several findings identified by static analysis tools, primarily focusing on optimizing parameter passing and minor code hygiene.

Key changes include:

*   **Optimized `QString` Parameter Passing**: Modified numerous function signatures across `Executor`, `MainWindow`, and `Util` classes to pass `QString` parameters by `const QString &` (constant reference) instead of by value. This change avoids unnecessary deep copies of `QString` objects, improving performance and efficiency.
*   **Refactored `Executor::executeBlocking` Call**: Adjusted an internal call within `Executor::executeBlocking` to pass the `app` parameter directly instead of using `std::move`, consistent with the updated `const QString &` signature.
*   **Minor Lambda Capture Optimization**: Removed an unnecessary `[this]` capture in a lambda connected to `MainWindow::passGitInitNeeded` in `qtpass.cpp`, as the `this` pointer was not utilized within the lambda's body.

## Summary by CodeRabbit

* **Refactor**
  * Optimized parameter passing for improved performance across multiple functions by reducing unnecessary data copies during function calls.
  * Removed unused variable capture in event handler callback.

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -220,7 +220,7 @@ static auto decodeAssumingUtf8(const QByteArray &in) -> QString {
  * Note: Returning error code instead of throwing to maintain compatibility
  * with the existing error handling pattern used throughout QtPass.
  */
-auto Executor::executeBlocking(QString app, const QStringList &args,
+auto Executor::executeBlocking(const QString &app, const QStringList &args,
                                const QString &input, QString *process_out,
                                QString *process_err) -> int {
   QProcess internal;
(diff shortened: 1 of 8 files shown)


nogeenharrie:
## Review: ✅ Approved

Reviewed all 8 files changed in this PR. All fixes correctly address the issues from #1026:

### Performance fixes (all correct)
- `executor.h/cpp` - Changed `QString app` to `const QString &` in 3 overloads
- `mainwindow.h/cpp` - Changed `QString dir` to `const QString &`  
- `util.h/cpp` - Changed `QString binary` to `const QString &`
- `gpgkeystate.cpp` - Changed `const QString record_type` to `const QString &`

### Bug fix (correct)
- `qtpass.cpp` - Removed unused `[this]` lambda capture

### Question
The `std::move(app)` was correctly removed in the second overload since you can't move a const reference - but did you verify this doesn't change behavior? The original passed ownership semantics, now it passes by const ref. Is that intentional for this call path?

---
**CI Status:** CodeFactor ✅, Hound ✅, License Compliance ✅, publiccode ✅ - waiting on rest

PR #1084, opened 2026-04-19
fix: guard against out-of-bounds index in password character set selection

annejan:
Adds a bounds check to the `on_createPasswordButton_clicked` function to prevent potential crashes.

The change validates the index retrieved from `ui->passwordTemplateSwitch` before using it to access `m_passConfig.Characters`. If the index is out of bounds, the password generation process is halted, and the UI is re-enabled, ensuring the application handles unexpected or invalid selections gracefully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to the password generation dialog to ensure character set selections are properly verified before generating passwords, preventing potential errors from invalid selections.

### DIFF
--- a/src/passworddialog.cpp
+++ b/src/passworddialog.cpp
@@ -94,10 +94,17 @@ void PasswordDialog::on_checkBoxShow_stateChanged(int arg1) {
  */
 void PasswordDialog::on_createPasswordButton_clicked() {
   ui->widget->setEnabled(false);
+  const int currentIndex = ui->passwordTemplateSwitch->currentIndex();
+  if (currentIndex < 0 ||
+      currentIndex >= static_cast<int>(m_passConfig.Characters.size())) {
+    ui->widget->setEnabled(true);
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


annejan:
### Code review

Found 1 issue:

1. `m_passConfig.Characters.size()` does not compile — `Characters` is a C-style array (`QString Characters[CHARSETS_COUNT]`), not a `std::array` or container, so it has no `.size()` member. The correct bound is `PasswordConfiguration::CHARSETS_COUNT`.

https://github.com/IJHack/QtPass/blob/9a1a62522b9a8af78fa1a8e0f3095df6ee1a2012/src/passworddialog.cpp#L96-L103

Fix already pushed to this branch: replaced `.size()` with `static_cast<int>(PasswordConfiguration::CHARSETS_COUNT)`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1129, opened 2026-04-21
Fix: QProgressIndicator memory leak and dark mode support

annejan:
Corrects the alpha calculation for the `QProgressIndicator` segments. An unnecessary `int` cast was truncating the floating-point alpha value, which likely resulted in an abrupt or incorrect visual fading effect. Removing the `int` cast ensures that the alpha value is applied as a floating-point number, providing a smoother and more accurate visual fade for the indicator.

## Release Notes

* **Bug Fixes**
  * Progress indicator now correctly adapts to the application's active color palette instead of defaulting to black when custom colors are not explicitly configured, ensuring improved visual consistency across different themes.

* **Other Changes**
  * Enhanced progress indicator implementation for better resource management.

### DIFF
--- a/src/keygendialog.cpp
+++ b/src/keygendialog.cpp
@@ -19,7 +19,7 @@
  * @param parent
  */
 KeygenDialog::KeygenDialog(ConfigDialog *parent)
-    : QDialog(parent), ui(new Ui::KeygenDialog) {
+    : QDialog(parent), ui(new Ui::KeygenDialog), m_progressIndicator(nullptr) {
   ui->setupUi(this);
   dialog = parent;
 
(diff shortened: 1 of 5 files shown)


annejan:
### Code review

Found 1 issue:

1. `qProgressIndicatorDisplayedWhenStopped` in `tst_util.cpp` is functionally identical to the existing `progressIndicatorSetDisplayedWhenStopped` in `tst_ui.cpp` — same assertions, same logic, only variable name differs. The new test should be removed and the existing one in `tst_ui.cpp` extended if additional coverage is needed.

New (duplicate) test:
https://github.com/IJHack/QtPass/blob/981b0924b1cd848645a8ee8178739d762ae289b0/tests/auto/util/tst_util.cpp#L991-L999

Existing equivalent test:
https://github.com/IJHack/QtPass/blob/981b0924b1cd848645a8ee8178739d762ae289b0/tests/auto/ui/tst_ui.cpp#L439-L446

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1130, opened 2026-04-22
fix: SimpleTransaction transactionEnd always queues item (#1130)

annejan:
Refactors the `transactionEnd` method in `simpleTransaction` to modify how `lastInTransaction` is handled and when items are enqueued.

Specifically, the changes ensure that:
*   `lastInTransaction` is now consistently reset to `INVALID` whenever `transactionEnd` is called (provided `transactionDepth` is greater than 0), rather than only when the transaction depth reaches zero.
*   An item is enqueued into `transactionQueue` if `lastInTransaction` is valid, removing the previous condition that `transactionDepth` must become zero for the enqueue operation to occur.

This adjustment aims to improve the reliability of transaction finalization, ensuring `lastInTransaction` is cleared and items are enqueued more consistently across nested transaction scopes.

### DIFF
--- a/src/simpletransaction.cpp
+++ b/src/simpletransaction.cpp
@@ -47,11 +47,11 @@ void simpleTransaction::transactionEnd(PROCESS pid) {
   dbg() << "END" << transactionDepth;
 #endif
   if (transactionDepth > 0) {
-    transactionDepth--;
-    if (transactionDepth == 0 && lastInTransaction != INVALID) {
+    if (lastInTransaction != INVALID) {
       transactionQueue.emplace(lastInTransaction, pid);
-      lastInTransaction = INVALID;
     }
+    transactionDepth--;
+    lastInTransaction = INVALID;
   }
 }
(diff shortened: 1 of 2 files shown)


nogeenharrie:
Closing - the AI-added source code change breaks nested transaction atomicity.

**Why the change is incorrect:**

The modified  would queue  every time it's called (when depth > 0), instead of only when depth reaches 0.

For nested transactions like:
# 0 "<stdin>"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 0 "<command-line>" 2
# 1 "<stdin>"

**New behavior (incorrect):** B is queued immediately when depth is still 1, breaking atomicity of the outer transaction.

**Original behavior (correct):** B waits until depth reaches 0, preserving atomicity of all nested operations.

The AI-added tests  and  relied on this incorrect behavior and should also be removed.

Issue #422, opened 2018-09-30
Simple way to create shared folder between multiple users in QtPass


MurzNN:
At now it's too hard to create shared with other users `pass` folder via QtPass UI: there are no ways to export current user public gpg id, import other users public gpg ids to folder, re-encrypt passwords after adding new user, etc.

Please make "Share" menuitem on folders, that provide dialog for export your public gpg id, import other users public gpg ids, re-encrypt passwords, and short text hints why we need to do this.

nogeenharrie:
## Implementation Plan

### Overview
Add a "Share" menu to folders in QtPass that enables sharing password folders with other GPG users.

### Current State
- Right-click on folder already has: Rename folder, Delete, Re-encrypt
- Re-encrypt exists but is hidden in context menu

### Proposed Implementation

Add a "Share" submenu with these options:

1. **Export My Public Key** - Export user's public GPG key to clipboard or file
2. **Add Recipient** - Import another user's public GPG key to folder
3. **Re-encrypt** (already exists, move to Share menu)
4. **Help** - Brief explanation of GPG sharing workflow

### Files to Modify

| File | Changes |
|------|---------|
| `src/mainwindow.cpp` | Add Share menu to contextMenu, new dialogs |
| `src/mainwindow.h` | Add slots for Share actions |
| `src/configdialog.cpp/h` | Add key import dialog |
| `src/imitatepass.cpp` | Use reencryptPath for bulk re-encryption |

### UI Flow

Right-click folder → Share submenu:
```
Share →
  ├── Export My Public Key...
  ├── Add Recipient...
  ├── Re-encrypt all passwords
  └── What is this? (opens help)
```

### Dialogs

1. **Export Public Key Dialog**
   - Show user's key fingerprint
   - "Copy to Clipboard" button
   - "Save to File" button

2. **Add Recipient Dialog**
   - Text input for key or file picker
   - Preview recipient name
   - "Add" button

3. **Help Dialog** (simple message box)
   - Explain GPG sharing workflow
   - Link to FAQ

### Implementation Steps

1. Add Share menu to folder context menu (mainwindow.cpp)
2. Create ExportPublicKeyDialog
3. Create AddRecipientDialog  
4. Wire up re-encrypt to new Share menu
5. Add documentation/help text

### Priority
Medium - Nice to have feature for team use.

annejan:
Closing this — the long-running umbrella request from 2018 has now landed in `main` across three PRs:

[...]

Issue #908, opened 2026-04-07
QtPass 2.0: Qt5 compatibility cleanup


nogeenhenk:
# QtPass 2.0: Qt5 Compatibility Cleanup

> **Status:** the `src/`, `main/` and CI items are done in #1679 on the `2.0-nazomer` branch (Qt 6.2 floor enforced in `qtpass.pri`). Line numbers below are from April and have drifted; the checkboxes are what count. Remaining: the `tst_locale` `QCOMPARE_NE` swap, which needs Qt 6.4 — do it if/when the floor moves past 6.2.

This issue tracks all code that can be simplified or removed once Qt5 support is dropped.

### src/pass.cpp
- [x] Line 288: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Lines 239-289: `splitCommandCompat()` function - no longer needed
- [x] Line 356: `AS_CONST()` macro usage

### src/gpgkeystate.cpp
- [x] Line 107: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Line 112: `QString::SkipEmptyParts` → `Qt::SkipEmptyParts`

### src/imitatepass.cpp
- [x] Line 297: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Line 349: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Line 451: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Lines 302, 354: `QString::SkipEmptyParts` → `Qt::SkipEmptyParts`

### src/keygendialog.cpp
- [x] Line 103: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Line 128: `#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)` - remove Qt5 fallback
- [x] Lines 108, 133: `QString::SkipEmptyParts` → `Qt::SkipEmptyParts`

### src/passworddialog.cpp
- [x] Lines 141, 147: `AS_CONST()` macro usage

### src/filecontent.cpp
- [x] Line 35: `AS_CONST()` macro usage

## Qt6-Only Code to Keep

These Qt6 features are already guarded and should remain:

[...]

PR #1162, opened 2026-04-25
feat: wire Share submenu actions to real operations (#422)

annejan:
This pull request introduces functional operations for sharing-related actions within the application, replacing previous instructional message boxes with interactive user interfaces.

Key changes include:
*   **Implemented Public Key Export:** A new `ExportPublicKeyDialog` has been added, allowing users to view their ASCII-armored public GPG key. This dialog provides options to copy the key to the clipboard or save it to a file. The `MainWindow::exportPublicKey` function now programmatically exports the configured GPG signing key using `gpg` and displays it in this new dialog. If no key is configured or the export fails, appropriate guidance or error messages are shown.
*   **Enhanced Recipient Management:** The `MainWindow::addRecipient` function has been updated to open the `UsersDialog` for managing recipients, providing a graphical interface for this task instead of displaying manual instructions.

## Summary by CodeRabbit

* **New Features**
  * Added an Export Public Key dialog to view an ASCII‑armored signing key, copy to clipboard (temporary “Copied!” feedback), or save to file with a sensible default filename and save dialog.
  * Improved recipient management: adding a recipient opens a dedicated dialog for easier folder selection.

* **Bug Fixes**
  * Informative warnings when no signing key is configured or when export fails or yields no output.

* **Tests**
  * Automated tests for filename sanitization, dialog UI, clipboard copy behavior, and save workflow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ tests/auto/simpletransaction/tst_simpletransaction
 tests/auto/filecontent/tst_filecontent
 tests/auto/gpgkeystate/tst_gpgkeystate
 tests/auto/integration/tst_integration
+tests/auto/exportpublickeydialog/tst_exportpublickeydialog
 **/*.gc*
 *.bak
 README.c*
--- a/src/exportpublickeydialog.cpp
+++ b/src/exportpublickeydialog.cpp
@@ -8,6 +8,7 @@
 #include <QFileDialog>
 #include <QFontDatabase>
 #include <QMessageBox>
+#include <QRegularExpression>
 #include <QSaveFile>
 #include <QTextStream>
 #include <QTimer>
@@ -54,9 +55,15 @@ void ExportPublicKeyDialog::on_copyButton_clicked() {
  *        destination and write the armored key to that file.
  */
 void ExportPublicKeyDialog::on_saveButton_clicked() {
-  QString defaultName = m_keyId.isEmpty()
+  // m_keyId may hold space-separated key IDs and is settings-controlled, so
+  // strip it down to the first whitespace token and keep only filename-safe
+  // characters before suggesting it as the default save name.
+  QString safeKeyId = m_keyId.section(QChar(' '), 0, 0);
+  static const QRegularExpression unsafeChars(QStringLiteral("[^A-Za-z0-9_-]"));
+  safeKeyId.remove(unsafeChars);
+  QString defaultName = safeKeyId.isEmpty()
                             ? QStringLiteral("public_key.asc")
-                            : QStringLiteral("%1.asc").arg(m_keyId);
+                            : QStringLiteral("%1.asc").arg(safeKeyId);
   QString fileName = QFileDialog::getSaveFileName(
       this, tr("Save Public Key"), defaultName,
       tr("ASCII-armored key (*.asc);;All files (*)"));
--- a/src/exportpublickeydialog.cpp
+++ b/src/exportpublickeydialog.cpp
@@ -1,15 +1,16 @@
-// SPDX-FileCopyrightText: YYYY Anne Jan Brouwer
+// SPDX-FileCopyrightText: 2026 Anne Jan Brouwer
 // SPDX-License-Identifier: GPL-3.0-or-later
 #include "exportpublickeydialog.h"
 #include "ui_exportpublickeydialog.h"
 
 #include <QApplication>
 #include <QClipboard>
-#include <QFile>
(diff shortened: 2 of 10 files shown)


annejan:
Review feedback addressed in fa77fa9:

- **SPDX year (`YYYY` → `2026`)** — applied. Matches the rest of the repo (51+ files use real years; only the two new files were using `YYYY`).
- **Multi-key identity** — fixed. `getPassSigningKey()` is now split on whitespace before being passed to `gpg --export`, mirroring the existing pattern in `imitatepass.cpp:373,425` and `pass.cpp:802`. Used `Qt::SkipEmptyParts` directly without a `#if QT_VERSION` guard since Qt 5.15 (the project minimum) supports it natively, and #908 tracks removing existing version guards.
- **`QSaveFile` + write error check** — applied. Atomic write via `QSaveFile`, checked `QTextStream::status()` and `file.commit()`, and surfaced any failure with `file.errorString()` in a `QMessageBox::warning`.
- **Copy button feedback** — applied. Relabels the button to `Copied!` for 1.5s via `QTimer::singleShot`, then reverts to the original (translated) text captured at construction.

Build clean (0 warnings/errors), Doxygen clean, clang-format clean, `make check` green (19/19 integration tests pass).

annejan:
Addressed in 8db353b + 2708502:

**Inline review items**

- **Sanitize `m_keyId` before using as filename** — applied. Extracted to a public static helper `ExportPublicKeyDialog::sanitizeKeyIdForFilename()` that takes the first whitespace-separated token and strips anything outside `[A-Za-z0-9_-]`. Needed since fa77fa9 allows `getPassSigningKey()` to be space-separated multi-key.
- **`keyIdLabel` rich-text injection** — fixed in the `.ui` by setting `textFormat = Qt::PlainText`, matching the property `UsersDialog`'s label already uses. The armored key text widget was already `QPlainTextEdit` so it never had this issue; only the label needed it.

**Coverage**

Added `tests/auto/exportpublickeydialog/` with 13 cases (8 data-driven sanitization rows + label-plain-text + plainTextEdit content/read-only + clipboard/relabel/revert). All pass locally; should bring patch coverage from 0% up.

**Duplicate finding — `QT_VERSION` guard around `Qt::SkipEmptyParts`**

I'd like to push back on this one. The technical premise — \"so the code compiles on older Qt versions\" — doesn't hold for this codebase:

- `Qt::SkipEmptyParts` was added in **Qt 5.14**.
- The project's minimum is **Qt 5.15** (`ccpp.yml` matrix, `qtpass.pri`), so on every supported Qt version `Qt::SkipEmptyParts` is the natively-available form.
- The existing guards in `imitatepass.cpp:373/425` and `pass.cpp:802` therefore always take the \`#if\` branch — the \`#else QString::SkipEmptyParts\` block is unreachable on every supported configuration.
- **Issue #908 explicitly tracks removing those guards** (\"QtPass 2.0: Qt5 compatibility cleanup\" lists them by file/line). Adding a new one here would create more work for that cleanup, not less.

Happy to add the guard if you'd rather have strict pattern parity now and clean them all up under #908; just say the word. But on the merits I think the cleaner code is correct, and `Qt::SkipEmptyParts` is the form #908 wants the codebase to settle on.

annejan:
Addressed in daaf32e:

**Whitespace handling in `sanitizeKeyIdForFilename`** — applied. The function now splits on `\\s+` with `Qt::SkipEmptyParts` and takes the first surviving token, so leading whitespace, tabs, and newlines are all normalized the same as spaces. Test cases updated:

- \`leading-whitespace\` (\`\"   DEADBEEF\"\` → \`\"DEADBEEF\"\`) — was previously expected to yield empty; now non-empty as the reviewer suggested.
- \`tab-separated\` (\`\"DEADBEEF\\tFEEDFACE\"\` → \`\"DEADBEEF\"\`) — added.
- \`only-whitespace\` (\`\"  \\t\\n \"\` → empty) — added.

All 15 cases pass; clang-format / Doxygen / build all clean.

**Duplicate `NSPrincipalClass` in the test plist** — applied to my new file (`tests/auto/exportpublickeydialog/qtpass.plist`). Quick context for the reviewer: I copied the plist from `tests/auto/ui/qtpass.plist`, which has the same duplicate, as does every other `tests/auto/*/qtpass.plist` in the repo (8 more files). CI tolerates it (every existing test passes today) but it's still redundant. Want me to fold a sweep cleanup of the other 8 plists into this PR, or open a separate `chore:` PR for that?

Issue #1167, opened 2026-04-25
Add 'Import key...' to UsersDialog so foreign keys can be added without leaving the share flow


annejan:
## Background

#1162 wired the Share → **Add Recipient…** action to the existing `UsersDialog`, so users can tick keys in their keyring as recipients for a folder's `.gpg-id`. What it does **not** yet handle is the moment a user wants to share with someone whose public key isn't in their keyring — they currently have to leave QtPass, run `gpg --import …`, then come back. That's the last unimplemented piece of the original Share-folder request (#422).

## Proposal

Add an **Import key…** button to `UsersDialog`, next to the existing search field.

Why there and not in Settings → GPG:

- It's the exact place where the gap surfaces. The user is trying to tick "Bob" as a recipient, can't find Bob, and needs an escape hatch right there — not three modals away.
- `UsersDialog` is reachable from both the toolbar **Users** button and from Share → **Add recipient…**, so a single addition benefits both flows.
- The keyring is global, but the keyboard-distance to import should be short from the per-folder dialog where the missing-recipient problem is visible.

(We can still add a parallel entry point in Settings → GPG later for completeness; the dialog button is the primary surface because that's where the friction is.)

## UI flow

1. User opens `UsersDialog` (toolbar Users button or Share → Add recipient…).
2. Searches for the new teammate; nothing matches.
3. Clicks **Import key…**.
4. Sub-dialog offers two paths:
   - **From file…** — `QFileDialog::getOpenFileName` for a `.asc` / `.gpg` file.
   - **From clipboard / paste…** — multiline text box pre-filled with clipboard if it looks like an armored block.
5. Run `gpg --import` via `Executor::executeBlocking` with the chosen input.
6. On success, parse `gpg --import`'s status output to extract the imported key fingerprint, refresh `loadGpgKeys()` / `populateList()`, and select the newly added entry so the user can tick it immediately.
7. On failure, show a `QMessageBox::warning` with `gpg`'s stderr.

## Files likely to change

| File | Change |
|------|--------|
| `src/usersdialog.{h,cpp,ui}` | Add **Import key…** button + slot |
| `src/importkeydialog.{h,cpp,ui}` (new) | File / clipboard import sub-dialog |
| `src/src.pro` | Register new dialog |
| `tests/auto/importkeydialog/` (new) | Unit tests (input parsing, sanitization, error paths) |

## Out of scope

- Editing/deleting keys (gpg's keyring management is its own UI surface).
[...]

PR #1170, opened 2026-04-25
feat: add Import key dialog to UsersDialog for issue #1167

nogeenharrie:
This pull request introduces a new "Import GPG Key" dialog, enhancing QtPass's key management capabilities.

Key changes include:
*   **New Import GPG Key Dialog:** A dedicated dialog (`ImportKeyDialog`) is added, allowing users to import GPG public keys.
    *   Users can choose to import a key from a file (supporting `.asc`, `.gpg`, `.key` extensions) or by pasting ASCII-armored key text directly into an input field.
    *   The dialog executes the `gpg --import` command internally and parses its output to confirm successful import and extract the key's fingerprint.
    *   It provides user feedback through success and error messages.
*   **Integration into Users Dialog:** A new "Import key..." button is added to the "Users" dialog.
    *   Clicking this button opens the new "Import GPG Key" dialog.
    *   Upon successful key import, the "Users" dialog automatically refreshes its list of GPG keys and attempts to select the newly imported key, making it immediately visible to the user.

This feature simplifies the process of adding GPG public keys to QtPass, directly addressing issue #1167 by providing a convenient in-application method for key import.

## Summary by CodeRabbit

* **New Features**
  * Adds an "Import key..." action to the Users dialog and a modal to import GPG public keys from file, clipboard, or pasted ASCII-armored text; import enabled only with non-empty input.

* **User Feedback**
  * Validates input, shows descriptive error dialogs on failure, and a success confirmation exposing the imported key id; newly imported key is auto-selected in the list.

* **Tests**
  * Adds automated tests for import-output parsing and dialog UI/behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ tests/auto/filecontent/tst_filecontent
 tests/auto/gpgkeystate/tst_gpgkeystate
 tests/auto/integration/tst_integration
 tests/auto/exportpublickeydialog/tst_exportpublickeydialog
+tests/auto/importkeydialog/tst_importkeydialog
 **/*.gc*
 *.bak
 README.c*
--- a/src/importkeydialog.cpp
+++ b/src/importkeydialog.cpp
@@ -134,7 +134,11 @@ bool ImportKeyDialog::importFromString(const QString &input) {
 }
 
 auto ImportKeyDialog::parseGpgImportOutput(const QString &output) -> QString {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
   const QStringList lines = output.split('\n', Qt::SkipEmptyParts);
+#else
+  const QStringList lines = output.split('\n', QString::SkipEmptyParts);
+#endif
   // Prefer locale-independent status output. IMPORT_OK gives a full
   // fingerprint; IMPORTED gives a (long) key id. Try both before falling
   // back to the English-only human-readable line.
--- a/src/importkeydialog.cpp
+++ b/src/importkeydialog.cpp
@@ -15,31 +15,32 @@
 #include <QRegularExpression>
 #include <QString>
 
+// Locale-independent: gpg's machine-readable status output via --status-fd 1.
+// See doc/DETAILS in the GnuPG source for IMPORT_OK / IMPORTED grammar.
 static const QRegularExpression
-    KEY_IMPORTED_EXACT(QStringLiteral(R"(gpg: key ([0-9A-Fa-f]+): imported)"));
-static const QRegularExpression KEY_IMPORTED_GENERAL(
+    IMPORT_OK_RE(QStringLiteral(R"(\[GNUPG:\] IMPORT_OK \d+ ([0-9A-Fa-f]+))"));
+static const QRegularExpression
+    IMPORTED_RE(QStringLiteral(R"(\[GNUPG:\] IMPORTED ([0-9A-Fa-f]+))"));
+// Fallback for the human-readable (English-locale) line.
+static const QRegularExpression KEY_IMPORTED_FALLBACK(
     QStringLiteral(R"(gpg: key ([0-9A-Fa-f]+):.*imported)"));
 
 ImportKeyDialog::ImportKeyDialog(QWidget *parent)
     : QDialog(parent), ui(new Ui::ImportKeyDialog) {
   ui->setupUi(this);
-
-  ui->inputTextEdit->setPlaceholderText(
-      tr("Paste an ASCII-armored GPG key here..."));
-
   ui->importButton->setEnabled(false);
 }
 
 ImportKeyDialog::~ImportKeyDialog() = default;
 
-QString ImportKeyDialog::importedKeyFingerprint() const {
+auto ImportKeyDialog::importedKeyId() const -> QString {
   return m_importedKeyId;
 }
 
 void ImportKeyDialog::on_fileButton_clicked() {
   const QString fileName = QFileDialog::getOpenFileName(
       this, tr("Import GPG Key"), QString(),
-      tr("GPG Key Files") + " (*.asc *.key);;" + tr("All Files") + " (*)");
+      tr("ASCII-armored GPG key") + " (*.asc);;" + tr("All Files") + " (*)");
 
   if (fileName.isEmpty()) {
     return;
@@ -52,16 +53,33 @@ void ImportKeyDialog::on_fileButton_clicked() {
     return;
   }
 
-  QByteArray bytes = file.readAll();
+  const QByteArray bytes = file.readAll();
   file.close();
+
+  // Only ASCII-armored content is shown in the text edit; binary keyrings
+  // would lose bytes through UTF-8 conversion. Reject anything that doesn't
+  // start with the PGP armor header.
+  if (!bytes.trimmed().startsWith("-----BEGIN PGP")) {
+    QMessageBox::warning(
+        this, tr("Import Key"),
+        tr("%1 does not look like an ASCII-armored GPG key. Convert it with "
+           "<code>gpg --armor --export</code> first, or paste the armored "
+           "block via <b>From Clipboard</b>.")
+            .arg(fileName));
+    return;
+  }
 
   ui->inputTextEdit->setPlainText(QString::fromUtf8(bytes));
 }
 
 void ImportKeyDialog::on_pasteButton_clicked() {
-  QClipboard *clipboard = QApplication::clipboard();
+  const QClipboard *clipboard = QApplication::clipboard();
   const QString text = clipboard->text();
-
+  if (text.isEmpty()) {
+    // Don't wipe whatever the user already typed/loaded for an empty
+    // clipboard.
+    return;
+  }
   ui->inputTextEdit->setPlainText(text);
 }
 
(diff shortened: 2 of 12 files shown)


annejan (commented):
Built locally (clean), Doxygen clean, full test suite green. Reviewing for fitness against the #1167 plan and for correctness:

## Fit vs #1167 plan

| #1167 ask | This PR |
|---|---|
| **Import key…** button on `UsersDialog` next to the search field | ✅ |
| Reachable from both toolbar Users and Share → Add recipient | ✅ (both already use `UsersDialog`) |
| Sub-dialog with **From file…** / **From clipboard / paste…** | ✅ button-driven |
| Multiline text box pre-filled with clipboard if it looks like an armored block | 🟡 explicit click only — see UX note below |
| Run `gpg --import` via `Executor::executeBlocking` | ✅ with `--status-fd 1 --batch --yes` |
| Parse status output, extract fingerprint, refresh `loadGpgKeys()`, **select the new entry** | ✅ |
| Error path: `QMessageBox::warning` with stderr | ✅ |
| `tests/auto/importkeydialog/` (new) | ❌ **not present** |

## Blockers

1. **No tests.** Same coverage gap that bit #1162 — codecov will flag 0% on the new files. The plan in #1167 explicitly listed `tests/auto/importkeydialog/`, and `parseGpgImportOutput` is a `static` helper specifically designed to be testable. Suggested cases: matching `gpg: key XXX: imported`, `gpg: key XXX: 1 imported`, multi-line output where the imported line is not first, output with no `imported` keyword (returns empty), empty input, lower-case hex digits, leading/trailing whitespace.

2. **Binary `.gpg` keyring files are broken.** `on_fileButton_clicked` reads any picked file with `QFile::readAll` and pushes it through `QString::fromUtf8(bytes)` into the plain-text edit. For a binary `.gpg` keyring blob this loses bytes through UTF-8 validation. Either:
   - restrict the file picker filter to `*.asc` only (and update the PR description, which currently claims `.gpg` is supported but the actual filter is `*.asc *.key`), **or**
   - on file pick, send the raw bytes straight to `gpg --import`'s stdin without bouncing through the text widget.

3. **`gpgExe` empty fallback missing.** `getGpgExecutable()` returns empty when not configured, and `Executor::executeBlocking(\"\", …)` then fails opaquely. #1162 fell back to `\"gpg\"` for the export path; same fallback (and the same fix) should apply here.

## Correctness / cleanup

4. **Selection vs. filter race in `UsersDialog::on_importKeyButton_clicked`.** The flow saves the filter → clears → populates unfiltered → selects the new item → re-applies the filter. The re-apply happens *after* `setCurrentItem`, so if the filter doesn't match the newly imported key it gets hidden again with the selection on it. Simpler: leave the filter cleared (the user just imported and wants to see/tick the new key), or set the filter to something that will match the new entry first, then `setCurrentItem`.

5. **Naming inconsistency.** `importedKeyFingerprint()` returns the value the regex captured from `gpg: key XXX:`, which is the long key ID (not the fingerprint — the fingerprint comes from the `FPR` status record). The member variable is `m_importedKeyId`. Pick one and use it everywhere — `importedKeyId()` matches the actual data and the member name.

6. **Two redundant regexes.** `KEY_IMPORTED_GENERAL` (`gpg: key (HEX):.*imported`) is a strict superset of `KEY_IMPORTED_EXACT` (`gpg: key (HEX): imported`). Drop the EXACT version.

7. **Defensive check is dead code.** After `dialog.exec() == Accepted`, `m_importedKeyId` is guaranteed non-empty (it's only set right before `accept()`). The `if (fingerprint.isEmpty())` branch in `usersdialog.cpp` can't fire on the success path, and its error text (\"use the Import button to import a key\") is misleading because the user just used it. Either drop the branch or assert.

8. **Two `private slots:` sections in `usersdialog.h`.** The new slot can join the existing one above it.

9. **`.ui` files have indentation drift.** `src/usersdialog.ui` and `src/importkeydialog.ui` mix indentation widths and have closing tags landing at unexpected columns (e.g. `<connections>` at column 0 in `importkeydialog.ui`, the whole `<item>` block in `usersdialog.ui` reindented). Cosmetic but worth a pass through `xmllint --format` or Qt Designer save.

## UX nit

10. The plan suggested auto-detecting an armored block on the clipboard and pre-filling. The button-only approach works but a one-line check on dialog open — `if clipboard.text().contains(\"BEGIN PGP PUBLIC KEY BLOCK\")` then prefill — saves a click without adding much code.

## Net

The shape and integration are right; the missing test suite plus the binary-file UTF-8 handling and the `gpgExe` fallback are the things that need to land before merge. Happy to provide more detail on any item.

annejan (commented):
Re-reviewed force-push (`bfff1ab9` → `db49679b`). Local build, Doxygen, and `clang-format` are now clean.

## Addressed ✅

- **`gpgExe` empty fallback** in `importFromString` — fixed.
- **clang-format violations** in both files — fixed.

## Still blocking

1. **No tests on the new dialog.** Same as before — `tests/auto/importkeydialog/` is not present and codecov will flag 0% on the new files. `parseGpgImportOutput` is `static` and trivial to cover; the `importFromString` error paths can be covered with a stub `gpgExe` (e.g., `/bin/false`) without touching the real keyring.

2. **Binary `.gpg` / `.key` files still corrupt-on-import.** Unchanged: `on_fileButton_clicked` reads any picked file with `QFile::readAll` and pushes it through `QString::fromUtf8(bytes)` into the plain-text widget. The file picker filter still advertises `*.asc *.key`. ASCII-armored `.asc` is fine; binary `.key`/`.gpg` blobs lose bytes through UTF-8 validation. Either narrow the picker to `*.asc` (and update PR description) or feed the raw bytes directly to `gpg --import` stdin without the text-edit detour.

## Regression introduced by the refactor

3. **The `QSignalBlocker` "fix" makes UX worse, not better.** The original concern was: after refresh, restoring the filter could hide the just-selected new key. The new code wraps the filter restore in a `QSignalBlocker` so `populateList` doesn't fire on the `textChanged` signal. Net result: the search field **shows the filter text** but **the list is no longer filtered by it**. The user sees a stale filter that doesn't match the list contents — confusing and a real bug because subsequent edits to the lineEdit will append to a filter that wasn't actually applied.

   The right shape is to leave the filter cleared after import, since the user's immediate goal is to see and tick the just-imported key. If preserving filter state matters, restore it *and* re-apply it (let `populateList` run), then check whether the imported key is hidden by the active filter and surface that to the user (e.g., a status-bar message: "Imported key XXX (hidden by current filter)").

## Cleanup still outstanding

4. **`importedKeyFingerprint()` returns a key ID, not a fingerprint.** The regex captures from `gpg: key XXX:`, which is the long key ID. The member variable is `m_importedKeyId`. Rename the public method to `importedKeyId()` so the API matches the data and the field.
5. **Redundant `KEY_IMPORTED_EXACT` regex.** `KEY_IMPORTED_GENERAL` (`gpg: key (HEX):.*imported`) is a strict superset; drop the EXACT version.
6. **Dead defensive branch in `usersdialog.cpp`.** After `dialog.exec() == Accepted`, `m_importedKeyId` is non-empty by construction. The `if (fingerprint.isEmpty())` branch can't fire on the success path, and the user-facing error text (\"Please use the Import button to import a key\") doesn't fit either. Drop the branch or assert.
7. **Two `private slots:` sections** in `usersdialog.h` — consolidate.
8. **`.ui` indentation drift** in `usersdialog.ui` and `importkeydialog.ui` — cosmetic but worth a pass through Qt Designer or `xmllint --format`.
9. **`src/importkeydialog.cpp` is missing the trailing newline** (`\ No newline at end of file` in the diff). Most editors and lints will eventually flag this.

## Net

Of the three blockers from my prior review, one (gpgExe) is resolved. The two structural ones — tests and binary-file handling — are unchanged. The filter-restore change went the wrong direction; that's worth reverting before the cleanup pass. Happy to elaborate on any item or send a follow-up patch if useful.

annejan (commented):
Re-checked force-push `db49679` → `484ee21`. The only change is a one-line manual reindentation of the file-open error message:

```
-                         tr("Could not open file: %1").arg(fileName));
+                      tr("Could not open file: %1").arg(fileName));
```

This **introduces** a clang-format violation at `src/importkeydialog.cpp:50` that wasn't there before — running `clang-format --style=file --dry-run` on the file now fails. (Easiest fix: `clang-format --style=file -i src/importkeydialog.cpp` and re-push.)

None of the structural items from the previous two reviews are addressed in this revision: tests still missing, binary file UTF-8 issue still present, `QSignalBlocker` filter-restore still shows a filter that's not actually applied, naming/regex/dead-branch/two-slots-sections/.ui-indentation/trailing-newline all unchanged.

Happy to send a patch covering the test suite + binary-file fix + filter restore, if that helps move things along.

annejan:
## Security review (post `ebe5c61`)

Ran a focused security review of the diff introducing \`ImportKeyDialog\` and the \`UsersDialog\` integration. **No HIGH or MEDIUM severity vulnerabilities identified.**

Patterns considered and ruled out (for the record):

- **\`Executor::executeBlocking(gpgExe, args, input, …)\`** — all argv elements are static literals (\`--status-fd\`, \`1\`, \`--import\`, \`--batch\`, \`--yes\`); the key material is delivered via stdin, not argv. No command-injection surface.
- **\`gpgExe = \"gpg\"\` PATH fallback** — same pattern QtPass already uses everywhere else for gpg invocations; not a new attack surface.
- **\`QFileDialog::getOpenFileName\`** — file path is the user's own interactive selection; no privilege boundary crossed, no traversal vector.
- **\`QMessageBox::warning\` body containing \`.arg(fileName)\` with HTML tags** — Qt's HTML subset is display-only (no JS execution); a pathological filename could mangle layout, not execute code.
- **\`QString::fromUtf8(bytes)\`** — invalid UTF-8 yields replacement chars; no memory-safety issue.
- **\`parseGpgImportOutput\`** — regexes capture hex-only character classes; the result is used solely to drive list-view selection behind explicit bounds checks (\`idx < 0 || idx >= m_userList.size()\`).
- **Clipboard text** — placed into a \`QPlainTextEdit\` via \`setPlainText\` (plain text, no rich-text interpretation); only fed to gpg if the user explicitly clicks Import.

The new code uses static argv lists, stdin for untrusted content, bounds-checked list indices, and plain-text rendering throughout — i.e. it follows the safe patterns already established in the codebase. Nothing concrete to flag.

annejan:
On the suggestion to replace
```cpp
QString gpgExe = QtPassSettings::getGpgExecutable();
if (gpgExe.isEmpty()) {
  gpgExe = \"gpg\";
}
```
with \`QString gpgExe = QtPassSettings::getGpgExecutable(\"gpg\");\`:

I'd push back here. The two patterns are not equivalent. \`QtPassSettings::getGpgExecutable(const QString &defaultValue)\` is implemented as \`QSettings::value(key, defaultValue).toString()\` (\`qtpasssettings.cpp:523-527\`), and \`QSettings::value\` returns the default **only when the key is missing**, not when it's stored as an empty string.

QtPass's bootstrap at \`qtpasssettings.cpp:497-499\` writes whatever \`Util::findBinaryInPath(\"gpg2\")\` returns into the \`gpgExecutable\` setting, which is \`\"\"\` on machines without gpg2 on PATH at first launch. After that initial write the key is present-but-empty, so \`getGpgExecutable(\"gpg\")\` would return \`\"\"\` — the manual \`isEmpty()\` check is what catches this case and lets the call fall through to a PATH-resolved \`gpg\`. Mirrors what #1162 did for the export path; dropping it would silently invoke \`Executor::executeBlocking(\"\", …)\` for those users.

Leaving the explicit check in place.

Issue #252, opened 2016-12-08
Feature: Process output in a separate area at the mainwindow


YoshiMan:
The process output should be displayed in a separated area. We need a setting, where the user can select, if the process output should be displayed or shouldnt. There should be a **clear** button to reset the *process output area*. It should be **scrollable** and should **auto scroll** to the last input, except the user scrolled a litte bit upwards, the autoscrolling is disabled. New outputs are appended at the bottom. It should be a red colored output, when the process has a non-zero exitvalue or has written something onto the stderr (black otherwise).

Notice: 
- We need a limit of the output lines.
- Are there any sensitive data nobody should see?

**when** the user enables process output to be displayed.
**then** the area will be displayed

**when** the user disables process output to be displayed.
**then** the area will not be displayed

**given** the user enables process output to be displayed.
**when** the output area contains more then it could show at once.
**then** the area displays a scrollbar 

**given** the user enables process output to be displayed.
**given** the output area contains less then it could show at once.
**when** the user clicks on the *clear output button*.
**then** the area will be cleared out.

**given** the user enables process output to be displayed.
**given** the output area contains more then it could show at once.
**when** the user clicks on the *clear output button*.
**then** the area will be cleared out and the scrollbar disapears.

**given** the output area contains more then it could show at once.
**when** the user clicks on the *clear output button*.
**then** the area will be cleared out and the scrollbar disapears.

**given** the user enables process output to be displayed.
**when** the *show password process* returns.
**then** the output should be displayed in the area for the passwords and **not** in the separate area.

**given** the user disables process output to be displayed.
**when** the *show password process* returns.
**then** the output should be displayed in the area for the passwords and **not** in the separate area.

**given** the user enables process output to be displayed.
**when** any other process, e.g. git commit, pass insert, git cp, git pull returns.
**then** a consecutive number, the executed command with parameter and the output should be displayed in a separate area at the mainwindow at the bottom at itself. 

**given** the user disables process output to be displayed.
**when** any other process, e.g. git commit, pass insert, git cp, git pull returns.
**then** a consecutive number, the executed command with parameter and the output should **not** be displayed anywhere.

**given** the user enables process output to be displayed.
**when** any other process, e.g. git commit, pass insert, git cp, git pull returns with an zero exitvalue.
**then** the output text should be black.

**given** the user enables process output to be displayed.
**when** any other process, e.g. git commit, pass insert, git cp, git pull returns with an non-zero exitvalue or has written something onto the stderr.
[...]

PR #1172, opened 2026-04-25
feat: add process output area to mainwindow (#252)

nogeenhenk:
This pull request introduces a new "Process Output" area to the main window, providing users with real-time feedback on commands executed by QtPass.

Key changes include:
-   **Process Output Display:** A new collapsible widget is added to the main window to display the standard output and error messages from `pass` commands. Error messages are highlighted in red.
-   **Configurable Visibility:** A new "Show process output" option is added to the configuration dialog, allowing users to enable or disable the visibility of this output area.
-   **Dynamic Visibility:** The output area automatically becomes visible when new output is generated, if the setting is enabled.
-   **Auto-scrolling:** The output area automatically scrolls to the latest output, but pauses auto-scrolling if the user manually scrolls up.
-   **Output Limiting:** The output area retains a maximum of 1000 lines to prevent excessive memory usage.
-   **Clear Functionality:** A "Clear" button is included to easily clear the displayed output.

## Summary by CodeRabbit

* **New Features**
  * New process output panel showing stdout/stderr (stderr highlighted); per-line numbering and optional command-name prefixes.
  * Auto-scroll that pauses while the user scrolls and resumes at the bottom; panel auto-opens when a command runs.
  * Clear button to reset output and counters.
  * Persistent "Show process output" setting with a Settings checkbox.
  * Session output capped and filtered to exclude sensitive commands.

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -77,6 +77,8 @@ ConfigDialog::ConfigDialog(MainWindow *parent)
   ui->plainTextEditTemplate->setPlainText(QtPassSettings::getPassTemplate());
   ui->checkBoxTemplateAllFields->setChecked(
       QtPassSettings::isTemplateAllFields());
+  ui->checkBoxShowProcessOutput->setChecked(
+      QtPassSettings::isShowProcessOutput());
   ui->checkBoxAutoPull->setChecked(QtPassSettings::isAutoPull());
   ui->checkBoxAutoPush->setChecked(QtPassSettings::isAutoPush());
   ui->checkBoxAlwaysOnTop->setChecked(QtPassSettings::isAlwaysOnTop());
@@ -287,6 +289,8 @@ void ConfigDialog::on_accepted() {
   QtPassSettings::setPassTemplate(ui->plainTextEditTemplate->toPlainText());
   QtPassSettings::setTemplateAllFields(
       ui->checkBoxTemplateAllFields->isChecked());
+  QtPassSettings::setShowProcessOutput(
+      ui->checkBoxShowProcessOutput->isChecked());
   QtPassSettings::setAlwaysOnTop(ui->checkBoxAlwaysOnTop->isChecked());
 
   QtPassSettings::setVersion(VERSION);
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -232,22 +232,22 @@
              </property>
             </widget>
            </item>
-<item>
-             <widget class="QCheckBox" name="checkBoxNoLineWrapping">
-              <property name="text">
-               <string>No line wrapping</string>
-              </property>
-             </widget>
-            </item>
-            <item>
-             <widget class="QCheckBox" name="checkBoxShowProcessOutput">
-              <property name="text">
-               <string>Show process output</string>
-              </property>
-             </widget>
-            </item>
-            <item>
-             <spacer name="horizontalSpacer_9">
+           <item>
+            <widget class="QCheckBox" name="checkBoxNoLineWrapping">
+             <property name="text">
+              <string>No line wrapping</string>
+             </property>
+            </widget>
+           </item>
+           <item>
+            <widget class="QCheckBox" name="checkBoxShowProcessOutput">
+             <property name="text">
+              <string>Show process output</string>
+             </property>
+            </widget>
+           </item>
+           <item>
+            <spacer name="horizontalSpacer_9">
              <property name="orientation">
               <enum>Qt::Horizontal</enum>
              </property>
--- a/src/configdialog.ui
+++ b/src/configdialog.ui
@@ -232,15 +232,22 @@
              </property>
             </widget>
            </item>
-           <item>
-            <widget class="QCheckBox" name="checkBoxNoLineWrapping">
-             <property name="text">
-              <string>No line wrapping</string>
-             </property>
-            </widget>
-           </item>
-           <item>
-            <spacer name="horizontalSpacer_9">
+<item>
+             <widget class="QCheckBox" name="checkBoxNoLineWrapping">
+              <property name="text">
+               <string>No line wrapping</string>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <widget class="QCheckBox" name="checkBoxShowProcessOutput">
+              <property name="text">
+               <string>Show process output</string>
+              </property>
+             </widget>
+            </item>
+            <item>
+             <spacer name="horizontalSpacer_9">
              <property name="orientation">
               <enum>Qt::Horizontal</enum>
              </property>
--- a/src/enums.h
+++ b/src/enums.h
@@ -3,6 +3,8 @@
 #ifndef SRC_ENUMS_H_
 #define SRC_ENUMS_H_
 
+#include <QMetaType>
+
 /**
  * @namespace Enums
  * @brief Enumerators for QtPass configuration and runtime operations.
@@ -46,5 +48,7 @@ enum PROCESS {
 };
 
 } // namespace Enums
+
+Q_DECLARE_METATYPE(Enums::PROCESS)
 
 #endif // SRC_ENUMS_H_
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -125,8 +125,17 @@ MainWindow::MainWindow(const QString &searchText, QWidget *parent)
   initStatusBar();
 
   connect(QtPassSettings::getPass(), &Pass::finishedAnyWithPid, this,
-          [this](const QString &out, const QString &err, int pid) {
-            if (pid == Enums::PASS_SHOW || pid == Enums::PASS_OTP_GENERATE) {
+          [this](const QString &out, const QString &err, Enums::PROCESS pid) {
+            // Never route potentially-secret output through the panel:
+            // - PASS_SHOW / PASS_OTP_GENERATE go via dedicated signals to
+            //   the main text browser (which clears on a timer).
+            // - PASS_GREP returns lines from password files; #252 must
+            //   not leak those into a long-lived panel.
+            // - PASS_INSERT's stdin is the password; stdout normally
+            //   carries gpg/git progress only, but exclude defensively
+            //   in case a future code path uses --echo or similar.
+            if (pid == Enums::PASS_SHOW || pid == Enums::PASS_OTP_GENERATE ||
+                pid == Enums::PASS_GREP || pid == Enums::PASS_INSERT) {
               return;
             }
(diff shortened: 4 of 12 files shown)


nogeenharrie:
## Review Gaps

Two issues identified when verifying against original requirements:

### 1. Password commands still appear in process output

**Requirement:** "when the *show password process* returns...the output should be displayed in the area for the passwords and **not** in the separate area"

**Current state:** The PR connects to `finishedAny` signal which fires for ALL commands including `pass show`. This means password content may appear in the process output area, defeating the purpose.

**Fix:** Emit a new signal that only fires for non-password commands, or pass the process type to MainWindow so it can filter:
- Add `enum PROCESS_TYPE { PASSWORD_COMMAND, OTHER_COMMAND }` to categorize
- Connect specific signals (`finishedGitInit`, `finishedInsert`, etc.) instead of `finishedAny`
- Or add parameter to `finishedAny(PROCESS pid, const QString &out, const QString &err)` to know the command type

### 2. Executed command with parameters not shown

**Requirement:** "the executed command with parameter and the output should be displayed"

**Current state:** Only line number shown, not the actual command executed.

**Fix:** Pass the command string to the display method:
- In Pass::executeWrapper, store the current command
- Add signal parameter: `finishedAny(PROCESS pid, const QString &cmd, const QString &out, const QString &err)`
- Display as: "N: cmd -> output" or "N: cmd\noutput"

### Suggested Implementation

1. In `pass.h`, change signal to include command type:
```cpp
void finishedAny(const QString &out, const QString &err, PROCESS pid);
```

2. In `pass.cpp`, emit with process ID:
```cpp
void Pass::emitProcessFinishedSignal(PROCESS pid, const QString &out, const QString &err) {
    emit finishedAny(out, err, pid);
}
```

3. In `mainwindow.cpp`, filter by process type:
```cpp
connect(QtPassSettings::getPass(), &Pass::finishedAny, this,
    [this](const QString &out, const QString &err, PROCESS pid) {
        if (pid == PASS_SHOW || pid == PASS_OTP_GENERATE) {
            return; // Skip - already handled elsewhere
        }
        // Show command info
        QString cmd = getProcessName(pid);
        onProcessOutput(cmd + ": " + out, !err.isEmpty());
    });
```

These changes align the PR with the original issue requirements.

nogeenharrie:
## Review Gaps

Two issues identified when verifying against original requirements in #252:

### 1. Password commands still appear in process output

**Requirement:** "when the *show password process* returns...the output should be displayed in the area for the passwords and **not** in the separate area"

**Current state:** The PR connects to `finishedAny` signal which fires for ALL commands including `pass show`. This means password content may appear in the process output area.

**Fix:** Filter by process type in MainWindow:
- Check the process ID passed in `finishedAny(out, err, pid)` 
- Skip if `pid == PASS_SHOW || pid == PASS_OTP_GENERATE`
- These already have their own signals (`finishedShow`, `finishedOtpGenerate`) that route to the password display

### 2. Executed command with parameters not shown

**Requirement:** "the executed command with parameter and the output should be displayed"

**Current state:** Only line number shown in format "N: output", not the actual command.

**Fix:** Access process name via `Enums::PROCESS` enum:
- The enum already has names like `GIT_PULL`, `PASS_INSERT`, etc.
- Display as "N: git pull -> output" or include command args

### Verified Good ✓

- Toggle setting (checkbox)
- Clear button 
- Auto-scroll with manual scroll detection
- Red/black coloring based on stderr presence
- Line limit (1000)
- Multi-line handling (my fix)
- Config dialog toggle (my fix)

The PR is close to done - these 2 gaps just need addressing.

nogeenharrie:
Both gaps have been addressed in recent commits:

1. **Password commands filtered** ✓ -  and  now skipped in the handler (lines 129-131)

2. **Command displayed** ✓ - Added  that maps process ID to command names like "git pull", "pass insert", etc. (lines 1773-1804)

The PR now fully matches issue #252 requirements. Ready for merge.

nogeenharrie:
Both gaps have been addressed in recent commits:

1. **Password commands filtered** - PASS_SHOW and PASS_OTP_GENERATE now skipped in the handler

2. **Command displayed** - Added getProcessName(pid) that maps process ID to command names like "git pull", "pass insert", etc.

The PR now fully matches issue #252 requirements. Ready for merge.

annejan:
Full review of #1172 — built locally, all 12 test suites green (\`make check\`), Doxygen clean, clang-format clean.

## Issue #252 mapping

The 2016 spec asked for: a toggleable, scrollable, auto-scrolling, clearable, line-limited output panel that colours errors red, hides \`pass show\` output, and is settings-controlled. This PR delivers all of that:

- ✅ \`Show process output\` setting + dynamic visibility on \`config()\`
- ✅ Read-only \`QTextEdit\` with min 80 / max 150 px
- ✅ Auto-scroll with pause-when-user-scrolls-up (slider press + \`value < maximum\` check)
- ✅ \`Clear\` button + counter reset
- ✅ 1000-line cap via \`limitOutputLines()\`
- ✅ Red colouring on stderr / errors
- ✅ \`PASS_SHOW\` / \`PASS_OTP_GENERATE\` excluded
- ✅ Numbered output lines

## Security review

One real privacy issue and one defensive one — both addressed in 94eeeff:

- **\`PASS_GREP\` was being routed into the panel.** \`pass grep\` returns lines from password files (the very point of the command), and \`finishedAnyWithPid\` fires for it. With \`Show process output\` enabled, those lines would land in the persistent panel — a clear leak relative to #252's intent. Excluded.
- **\`PASS_INSERT\` defensively excluded** as well. Today's \`pass insert -m\` and \`gpg --encrypt --output\` paths don't echo the password to stdout/stderr, but a future code path with \`--echo\` or comparable would. Cheaper to exclude now than to discover later.
- **HTML rendering** is fine: \`line.toHtmlEscaped()\` is applied to the user content, the colour comes from \`QColor::name()\` (\`#RRGGBB\` only), and the line number is from \`QString::number(int)\`. No injection vector. ✅
- \`PASS_SHOW\` / \`PASS_OTP_GENERATE\` already correctly excluded. ✅

## Inline review findings

| Finding | Verified | Resolution |
|---|---|---|
| Skip \`PASS_GREP\` (and consider \`PASS_INSERT\`) in the lambda | ✅ Real privacy issue for grep | Both added to the early-return |
| \`getProcessName\` returns raw English literals — \`tr()\` or \`// no-tr\` | ✅ | Wrapped each in \`QStringLiteral\` with \`// no-tr\` comment; these reference shell commands and shouldn't be localized. Also added \`GIT_MOVE\` / \`GIT_COPY\` (previously hit the empty default) |
| Unused \`int pid\` on \`appendProcessOutput\`; \`getProcessName\` should be \`static\` | ✅ \`pid\` was never read inside \`appendProcessOutput\`; \`getProcessName\` touches no instance state | Removed the parameter; marked the helper \`static\`; switched it to take \`Enums::PROCESS\` |
| Retype \`finishedAnyWithPid\` to \`Enums::PROCESS\`; add \`Q_DECLARE_METATYPE\`; drop \`static_cast<int>\` | ✅ | Done in \`pass.h\`, \`pass.cpp\`, \`enums.h\`, and the receiver lambda; \`#include <QMetaType>\` added in \`enums.h\` |

## Smaller observations addressed

- The \`switch\` in \`getProcessName\` previously had a \`default:\` returning empty for the secret-bearing pids. Replaced with explicit cases for \`PASS_SHOW\`, \`PASS_OTP_GENERATE\`, \`PROCESS_COUNT\`, \`INVALID\` so the compiler will warn if a new \`Enums::PROCESS\` member is added without a label.
- \`mainwindow.cpp\` was missing a trailing newline. Added.

## Things I did **not** change (deliberately)

- \`processOutputWidget\` placement via \`statusBar()->addPermanentWidget\` — unusual UX (the status bar is normally one row tall) but a UX call, not a correctness issue. Worth a separate UX pass before release.
- Auto-scroll lock heuristic (\`m_autoScroll = sb->value() >= sb->maximum()\`) — works, but flips back to true the moment the user scrolls back to bottom even by one tick. Acceptable behaviour, just noting.

## State after the push

Build: 0 warnings / 0 errors. Doxygen: 0 warnings. clang-format: clean. \`make check\`: all suites green (\`tst_util\`, \`tst_ui\`, \`tst_settings\`, \`tst_passwordconfig\`, \`tst_filecontent\`, \`tst_storemodel\`, \`tst_simpletransaction\`, \`tst_gpgkeystate\`, \`tst_exportpublickeydialog\`, \`tst_importkeydialog\`, \`tst_executor\`, \`tst_integration\`).

annejan:
On the two follow-up findings:

**1. Per-line prefix on multi-line output** — already addressed in 9ec5a93. `onProcessOutput` is now a one-liner that forwards `getProcessName(pid)` as the optional `linePrefix` argument to `appendProcessOutput`, and the prefix is applied inside the per-line loop. Multi-line output (e.g. a `git push` with three progress lines) is now consistently labelled `git push: …` on every line. ✅

**2. `Enums::GIT_COPY` returns `"git cp"`** — pushing back here. The premise in the comment ("the file is copied with QFile::copy() and then gitCommit() stages/commits it") is a misreading of `imitatepass.cpp:933-956`:

```cpp
void ImitatePass::Copy(const QString src, const QString dest, const bool force) {
  transactionHelper trans(this, PASS_COPY);
  if (QtPassSettings::isUseGit()) {
    QStringList args;
    args << \"cp\";
    if (force) args << \"-f\";
    args << pgit(src) << pgit(dest);
    executeGit(GIT_COPY, args);          // <-- this is the GIT_COPY tag
    gitCommit(\"\", message);
  } else {
    QFile::copy(src, dest);               // <-- QFile::copy is in the else
  }
  ...
}
```

The `GIT_COPY` process tag is only emitted in the `isUseGit()` branch, which literally invokes `git cp <src> <dest>`. The `QFile::copy()` path lives in the `else` branch and never emits `GIT_COPY`. So the label *is* what was run.

Worth knowing: `git cp` isn't part of stock git, but it ships in **`git-extras`** (Debian/Ubuntu `git-extras`, Homebrew `git-extras`, Arch `git-extras`). On systems where the user *can* reproduce `git cp`, the label is correct and useful. Relabelling to `\"git add && git commit\"` would mislabel everywhere (that's not what's run) and would obscure the actual mechanism for someone debugging a failure (they'd never guess to install git-extras).

If we want to take a real swing at this, the right move is to replace the `executeGit(GIT_COPY, {\"cp\", …})` call with `QFile::copy` + `git add` + `git commit` so QtPass doesn't depend on a third-party git extension. That's a real bug fix, but it's its own PR — not in scope for #1172.

Leaving the label as-is.

Issue #1513, opened 2026-06-03
Refactor: tighten the Pass interface (executeWrapper, Move/Copy duplication, Grep regex semantics, backend factory)


annejan:
Part of #1508 (thermonuclear code-quality audit). Four related Pass-hierarchy cleanups bundled because they all touch the same files.

## 1. \`executeWrapper\` 5-fold dispatch

\`pass.h:235–269\`, \`pass.cpp:111–125\`, \`imitatepass.cpp:1040–1044\`:

- \`Pass::executeWrapper(5-arg)\` → delegates to \`Pass::executeWrapper(6-arg)\`
- \`Pass::executeWrapper(6-arg, virtual)\` → calls \`exec.execute()\`
- \`ImitatePass::executeWrapper(6-arg, override)\` → calls \`transactionAdd(id)\` then \`Pass::executeWrapper(6-arg)\`

Three dispatch sites where one would do. Collapse to a single virtual \`executeCommand()\`; surface the ImitatePass transaction wrapping as an explicit \`Pass::beforeExecute()\` / \`afterExecute()\` hook rather than hiding it in an override.

## 2. Move/Copy duplication

\`realpass.cpp:123–193\` and \`imitatepass.cpp:903–969\` have ~60 lines of nearly identical path-normalisation / force-flag / \`.gpg\`-suffix logic. Extract a static helper on \`Pass\`:

\`\`\`cpp
struct ResolvedPaths { QString normalizedSrc, normalizedDest; bool isDirMove; };
static ResolvedPaths Pass::normalizePaths(QString src, QString dest, bool force);
\`\`\`

Both backends become thin: call \`normalizePaths\`, then execute backend-specific subprocess.

## 3. Grep regex mismatch (Liskov violation)

\`RealPass::Grep\` (\`realpass.cpp:200–206\`) shells out to \`pass grep\` which uses POSIX BRE.
\`ImitatePass::Grep\` (\`imitatepass.cpp:1105–1174\`) uses \`QRegularExpression\` which is PCRE-like.

The same pattern produces a different match set depending on which backend the user has configured. The caller has no way to know.

Options:
- Document loudly in the Pass header and the user-facing search field.
- Translate via a \`Pass::normalizeGrepPattern(QString, BackendKind) -> QString\` helper for the BRE↔PCRE conversion subset users actually use (anchors, basic quantifiers, character classes).
- Or pick one engine for both backends (probably PCRE on top of \`pass show | grep -P\` for RealPass).

[...]

Issue #1508, opened 2026-06-03
Tracking: thermonuclear code-quality audit findings


annejan:
This issue tracks the findings from the codebase-wide thermonuclear code-quality audit run on 2026-05-18 (post #1488 merge). Four parallel sub-agents reviewed every \`src/*.{cpp,h}\` against an "ambitious / structural-simplification" rubric and the synthesis surfaced two real bugs and several large structural opportunities.

## Discipline

- Small number of high-conviction structural improvements, not a flood of nits.
- Each linked sub-issue is independently shippable.
- The two bugs are 1-PR fixes; the structural ones range from 1 PR to a multi-PR series.

### Bugs (small, ship first)

- [ ] #1509 — WSL \`pgit\`/\`pgpg\` path shims appear broken (do not shell-evaluate)
- [ ] #1510 — Small cleanups: unreachable \`cdUp\` guard + \`Executor::executeBlocking\` overload sprawl

### Structural (bigger, design-heavy)

- [ ] #1511 — Replace \`QtPassSettings\` get/set boilerplate with an \`AppSettings\` struct + serialiser (**~1050 lines deleted**)
- [x] #1512 — Decompose \`MainWindow\` (currently 2096 lines, 2× the limit) into \`PasswordDisplayPanel\` + \`GrepSearchController\` + \`UiState\`
- [x] #1513 — Tighten the Pass interface (\`executeWrapper\` indirection, \`Move\`/\`Copy\` duplication, \`Grep\` regex semantics, backend factory)
- [x] #1514 — Split \`Util\` grab-bag into focused modules + consolidate \`StoreModel\` drag-drop handlers

## Out of scope here

Code that's already clean — recent dialog tests, drag-drop path-traversal hardening, default-args refactor (#1488), the 3 enabled Google clang-tidy checks (#1487). Don't re-audit those.

## Sequencing

Two reasonable tracks:

**Track A — bugs + small wins first, then structural**: ship #1509, #1510 first. Then tackle #1513 and #1514 in parallel (independent). #1511 and #1512 are big design commitments — do them when there's bandwidth for the design discussion.

**Track B — biggest wins first**: #1511 (settings model) first; everything downstream becomes easier (#1513 part 4 needs it; #1512 benefits from it).

Recommend Track A unless we commit to a focused refactor sprint.

## Notes

[...]

PR #1522, opened 2026-06-13
refactor: dedup RealPass::Move/Copy via passMoveOrCopy (#1513)

annejan:
Part of #1508 / #1513.

## What

`RealPass::Move` and `RealPass::Copy` were byte-identical except the pass subcommand (`mv`/`cp`) and process id (`PASS_MOVE`/`PASS_COPY`) — same force-skip guard, store-relative path mapping, `.gpg`-suffix stripping, and arg assembly. Extracted a private `passMoveOrCopy(id, subcommand, src, dest, force)`; `Move`/`Copy` are now one-line delegations.

`-33 / +30`.

## Scope note

Only the **RealPass internal** duplication is addressed. The cross-backend "normalizePaths in `Pass`" idea from #1513 doesn't fit — `ImitatePass::Move`/`Copy` are structurally different (git mv/cp + direct file ops + reencrypt) and don't share the store-relative/`.gpg` path shape. The `executeWrapper` indirection, the `Grep` BRE-vs-PCRE semantics question, and the backend-factory parts of #1513 remain open; **#1513 stays open**.

## Test plan

- [x] `qmake6 -r && make` clean
- [x] `tst_integration` 20/20 (incl. RealPass paths, real `pass`)
- [x] doxygen clean, clang-format applied

## Summary by CodeRabbit

* **Refactor**
  * Consolidated duplicate code in move and copy operations for improved maintainability.

### DIFF
--- a/src/realpass.cpp
+++ b/src/realpass.cpp
@@ -121,37 +121,7 @@ void RealPass::Init(QString path, const QList<UserInfo> &users) {
  * @param force overwrite
  */
 void RealPass::Move(const QString src, const QString dest, const bool force) {
-  QFileInfo srcFileInfo = QFileInfo(src);
-  QFileInfo destFileInfo = QFileInfo(dest);
-
-  // force mode?
-  // pass uses always the force mode, when call from eg. QT. so we have to
-  // check if this are to files and the user didn't want to move force
-  if (!force && srcFileInfo.isFile() && destFileInfo.isFile()) {
-    return;
-  }
-
-  QString passSrc = QDir(QtPassSettings::getPassStore())
-                        .relativeFilePath(QDir(src).absolutePath());
-  QString passDest = QDir(QtPassSettings::getPassStore())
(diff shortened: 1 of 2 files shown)


annejan:
Skipping the `QDir::cleanPath()` normalization suggestion:

- This block is **pre-existing** code that this PR only relocated (Move/Copy → `passMoveOrCopy`); the dedup didn't change the path logic.
- It's effectively already normalized: `QDir(src).absolutePath()` removes `.`/`..` and redundant separators internally, `QtPassSettings::getPassStore()` is normalized via `Util::normalizeFolderPath`, and `QDir::relativeFilePath()` cleans its inputs as well.
- No concrete failing input is given, and the RealPass move/copy path is exercised by the integration suite (passing).

Adding `cleanPath()` here would be redundant churn on behaviour-unchanged moved code. If there's a real reproducer (e.g. a store path with embedded `..` or doubled separators that misbehaves), happy to address it as its own fix.

Issue #1598, opened 2026-07-01
GPG keygen reports success unconditionally — real gpg errors are swallowed


annejan:
## Summary

When generating a GPG key pair, QtPass shows *"GPG key pair generated successfully"* even when `gpg` failed, and shows **no error** when it fails silently. There is no diagnostic path for a failing key generation.

## Where

`Pass::GenerateGPGKeys()` runs:

```
gpg --gen-key --no-tty --batch
```

On completion (`src/pass.cpp`):

```cpp
case GPG_GENKEYS:
  emit finishedGenerateGPGKeys(out, err);   // no exit-code awareness
  break;
```

Handler (`src/qtpass.cpp`, `onKeyGenerationComplete`):

```cpp
if (nullptr != m_mainWindow->getKeyGenDialog()) {
  m_mainWindow->cleanKeygenDialog();
  m_mainWindow->showStatusMessage(tr("GPG key pair generated successfully"), 10000);
}
processFinished(p_output, p_errout);
```

The success message is emitted purely on dialog presence — `p_errout` and the process exit code are never inspected. A non-zero `gpg` exit produces a false "success" (or nothing, if the dialog is already gone).

## Impact

- Users get no feedback when key generation fails (reported in #637, Flatpak sandbox where `gpg` couldn't reach its agent).
- Masks environment problems (missing agent, read-only `~/.gnupg`, entropy, etc.).

## Proposed fix

- Thread the exit code through to `onKeyGenerationComplete` (or check it in the wrapper) and only show the success message on exit 0.
- On non-zero exit or non-empty `stderr`, surface `stderr` via `critical()` / `flashText(..., true)` instead of the success message.

## Related

- #637 (Flatpak on Flathub) — this silent-success behaviour is why the keygen failure there produced "no output about why".

nogeenharrie:
Resolved.

**Original report** ("success shown regardless of exit code") was already fixed in `8df346373`: `Pass::finished` gates on `if (exitCode != 0) { handleProcessError(...); return; }` before any success signal, so "GPG key pair generated successfully" is only reachable on a clean exit.

[...]

Issue #637, opened 2023-07-01
Flatpak on Flathub


JBBgameich:
Thank you for this nice application :)

In order to make it easier to run on Fedora Kinoite / Silverblue, I'd like to submit it to Flathub.
Flathub requires that app developers agree to their apps being published there, and prefers if upstream developers also maintain the flatpak.

I'd offer to submit and maintain it there, but if you are interested in doing that, that would be even better of course :)

Please let me know if you have any questions about this.

annejan:
I'm very happy with you maintaining it, since I don't use flatpack myself . .

I am already FreeBSD maintainer and some more packages 😉

Justinzobel:
I saw this as I was curious about having this app on Flathub. Hi @jbruechert :D

I looked on Flathub but didn't see it on there, and no PR so I decided to start working on one. It is a bit hacky for now, but it opens, I need to add ssh, but ran out of time.

```
{
    "id": "org.qtpass.qtpass",
    "runtime": "org.kde.Platform",
    "runtime-version": "6.10",
    "sdk": "org.kde.Sdk",
    "command": "qtpass",
    "finish-args": [
        "--device=dri",
        "--share=ipc",
        "--socket=fallback-x11",
        "--socket=wayland",
        "--talk-name=org.kde.StatusNotifierWatcher"
    ],
    "modules": [
        {
            "name": "qtpass",
            "buildsystem": "simple",
            "build-commands": [
                "qmake6 PREFIX=/app",
                "make",
                "make DESTDIR=/app install",
                "install -Dm00644 qtpass.desktop /app/share/applications/org.qtpass.qtpass.desktop",
                "install -Dm00644 artwork/icon.svg /app/share/icons/hicolor/scalable/apps/org.qtpass.qtpass.svg",
                "sed -i 's/qtpass-icon/org.qtpass.qtpass/g' /app/share/applications/org.qtpass.qtpass.desktop"
            ],
            "sources": [
                {
                    "type": "archive",
                    "url": "https://github.com/IJHack/qtpass/archive/v1.7.0.tar.gz",
                    "sha256": "16d7bf01114a7f11c2e361b46e15063f419a27beff54a96985b706da756703c3",
                    "x-checker-data": {
                        "type": "anitya",
                        "project-id": 10288,
                        "stable-only": true,
                        "url-template": "https://github.com/IJHack/qtpass/archive/v${version}.tar.gz"
                    }
                }
            ]
        }
    ]
}
```

[...]

PR #1599, opened 2026-07-01
fix: report gpg keygen start failures instead of hanging (#1598)

nogeenharrie:
## Summary

Closes the residual silent-failure path behind #1598 (and the "no output about why" symptom in #637).

The original report — "success shown regardless of exit code" — was already fixed in `8df346373`: `Pass::finished` gates on `if (exitCode != 0) { handleProcessError(...); return; }` before any success signal, so `finishedGenerateGPGKeys` → "generated successfully" is only reachable on exit code 0. A keygen that *runs and returns non-zero* already shows "GPG key pair generation failed".

What was **not** covered: a gpg that **never starts**.

- Keygen sends its batch on stdin, so `Executor::executeNext()` takes the `waitForStarted(-1)` branch. On `FailedToStart` (missing/non-executable binary, or a Flatpak seccomp denial) it dequeued the command **without emitting `finished()` or `error()`** — so `Pass::finished` never ran, `cleanKeygenDialog()` never ran, and the modal keygen dialog spun forever with no feedback.
- An empty configured gpg executable hit the same dead-end via `Executor::execute()`'s intentional empty-app early return.

## Changes

- **`Executor::executeNext()`** — on `FailedToStart`, emit `error(id, -1, …, "Failed to start <app>")` so the failure routes through `Pass::finished`'s existing non-zero exit-code gate to the "generation failed" path. Emitted via a **queued** call so it does not re-enter `KeygenDialog::done()` (which drives key generation synchronously); this makes it behave exactly like the normal async `QProcess::finished` path. Fixes every stdin-carrying command, not just keygen.
- **`Pass::GenerateGPGKeys()`** — when the gpg executable is unconfigured, surface `processErrorExit()` instead of handing an empty command to the `Executor`. The Executor's silent empty-app return is left intact (it is intentional, to avoid spamming bogus failures for unconfigured git).
- **Regression tests** (`tests/auto/executor`):
  - `executeAsyncFailedToStartEmitsError` — a failed-to-start stdin command now emits exactly one `error()` with a non-zero code (previously **zero** signals).
  - `executeAsyncCrashExitReportsNonZeroCode` — locks in that a signal-killed process reports `CrashExit` with a non-zero exit code, so it can never be mistaken for success.

## Notes

- A signal-killed gpg was **verified** to already route correctly: on Unix (Linux verified; forkfd is identical on Qt 5.15) `QProcess` reports `CrashExit` with `exitCode == signal number` (SEGV→11, KILL→9…), never 0 — so it already hits the non-zero gate. No false-success there.
- Signal signatures are unchanged → Qt 5.15 and Qt 6 compatible.
- Deliberately **not** wiring `QProcess::errorOccurred` as the primary fix: for stdin-carrying commands it would double-emit with the `waitForStarted` branch.

## Testing

`make check` — all suites pass (executor: 34 including the 2 new tests; full run 0 failures).

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed-to-start processes now report an error (exit code `-1`) instead of silently stopping, preventing waits from hanging.
  * Key generation now validates GPG executable configuration and cleanly reports “No GPG executable configured” when missing.
  * Crash termination now consistently surfaces as an error with a non-zero exit code.
* **Tests**
  * Added async regression tests for failed start and crash behavior.
  * Added an integration test for key generation error handling with empty GPG executable.
  * Improved working-directory assertion to compare canonical paths.

### DIFF
--- a/src/executor.cpp
+++ b/src/executor.cpp
@@ -82,9 +82,26 @@ void Executor::executeNext() {
 #ifdef QT_DEBUG
           dbg() << "Process failed to start:" << i.id << " " << i.app;
 #endif
+          // Capture before dequeue() invalidates the head reference.
+          const int failedId = i.id;
+          const QString failedApp = i.app;
           m_process.closeWriteChannel();
           running = false;
           m_execQueue.dequeue();
+          // Surface the failure instead of silently dropping the command:
+          // otherwise callers waiting on a finished/error signal (e.g. the GPG
+          // keygen dialog) hang forever with no feedback. Defer the emit via a
+          // queued call so we do not re-enter a caller still on the stack —
+          // KeygenDialog::done() drives key generation synchronously — which
+          // mirrors the normal asynchronous QProcess::finished path. A -1 exit
+          // code routes through Pass::finished's non-zero error gate.
+          QMetaObject::invokeMethod(
+              this,
+              [this, failedId, failedApp]() {
+                emit error(failedId, -1, QString(),
+                           tr("Failed to start %1").arg(failedApp));
+              },
+              Qt::QueuedConnection);
           executeNext();
           return;
         }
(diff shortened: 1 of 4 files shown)


annejan:
@coderabbitai review

PR #1602, opened 2026-07-02
fix: stop ConfigDialog from silently corrupting saved settings

annejan:
## Summary

Batch 2 of the full `src/` review (see the audit): the config dialog silently overwrote or destroyed persisted settings on a normal **OK** click — no crash needed. This fixes the confirmed data-loss/corruption findings, all in `configdialog.cpp` + `qtpasssettings.cpp`.

## Fixes

**Password-generation settings wiped every OK** (HIGH)
`applySettings()` never loaded the pwgen path or password configuration, so `readSettings()` wrote the `.ui` defaults back over the user's saved length / charset / custom chars / usePwgen. Now restored via `setPwgenPath()` / `setPasswordConfiguration()` before `usePwgen()`.

**Custom charset clobbered** (part of the same finding)
`getPasswordConfiguration()` read the character line edit unconditionally, replacing the saved custom charset with a builtin string when a builtin was selected. Now only reads the line edit while `CUSTOM` is selected; otherwise preserves the persisted custom value.

**Profile table corruption + data loss** (HIGH)
`setProfiles()` populated the table with sorting enabled, so writing the sort column moved rows between `setItem()` calls — path/signing-key cells landed in the wrong profile's row and were destroyed on save (with a randomized `QHash` order, any user with 2+ profiles hit it intermittently). Now disables sorting during population and locates the current row by item pointer.

**Add-profile renames an existing profile** (MED)
`on_addButton_clicked()` re-enabled sorting then read `item(n, 0)`, which after the re-sort could be an existing profile — the edit/selection landed on the wrong row. Now tracks the new item by pointer.

**Null deref on double-click** (MED)
`on_profileTable_cellDoubleClicked()` dereferenced the path cell with no null check.

**OK enabled with empty required fields** (LOW)
`validate()` derived the button state from a single changed item, re-enabling OK while another row still had an empty name — allowing an empty-named profile to be saved and then silently dropped.

**Active profile wiped on save** (HIGH)
`QtPassSettings::setProfiles()` `remove()`d the overloaded `"profile"` key, which also stores the active-profile name (`getProfile`), silently switching the active password store on every save. Now preserves and restores the active-profile scalar.

## Tests

- `tst_settings::setProfilesPreservesActiveProfile` — active profile survives `setProfiles()`.
- `tst_configdialog::customCharsetRoundTrip` — CUSTOM charset survives set/get.
- Full suite: 22 binaries, 0 failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored saved pwgen path and password-generation settings when reopening the configuration dialog.
  * Improved profile-table behavior with sorting—new profiles are selected correctly after insertion and loaded for the right profile.
  * Made required-field validation more consistent, including OK button enablement across the whole table.
  * Prevented issues when double-clicking empty profile directory cells.
  * Preserved custom password character sets correctly when switching between custom and built-in options.
  * Ensured the active profile remains unchanged when updating saved profile settings.
* **Tests**
  * Added Qt coverage for custom charset round-trips and for adding/selecting new profiles when the table is sorted.

### DIFF
--- a/src/configdialog.cpp
+++ b/src/configdialog.cpp
@@ -1269,6 +1269,9 @@ void ConfigDialog::usePwgen(bool usePwgen) {
 
 void ConfigDialog::setPasswordConfiguration(
     const PasswordConfiguration &config) {
+  // Retain the custom charset so it survives even when a builtin set is the
+  // active selection (the line edit then shows the builtin's characters).
+  m_customPasswordChars = config.Characters[PasswordConfiguration::CUSTOM];
   ui->spinBoxPasswordLength->setValue(config.length);
   ui->passwordCharTemplateSelector->setCurrentIndex(config.selected);
   if (config.selected != PasswordConfiguration::CUSTOM) {
@@ -1285,14 +1288,12 @@ auto ConfigDialog::getPasswordConfiguration() -> PasswordConfiguration {
   // The line edit only holds the user's custom charset while CUSTOM is
   // selected; for a builtin selection it shows that builtin's characters.
   // Reading it unconditionally would overwrite the saved custom charset with a
-  // builtin string, so preserve the persisted custom value in that case.
+  // builtin string, so fall back to the retained custom value in that case.
   if (config.selected == PasswordConfiguration::CUSTOM) {
     config.Characters[PasswordConfiguration::CUSTOM] =
         ui->lineEditPasswordChars->text();
   } else {
-    config.Characters[PasswordConfiguration::CUSTOM] =
-        QtPassSettings::getPasswordConfiguration()
-            .Characters[PasswordConfiguration::CUSTOM];
+    config.Characters[PasswordConfiguration::CUSTOM] = m_customPasswordChars;
   }
   return config;
 }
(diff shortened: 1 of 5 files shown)


annejan:
Good catch — addressed in 1984c2c74. Added `customCharsetPreservedWhenBuiltinSelected`, which sets a CUSTOM charset, switches the selection to a builtin (ALPHANUMERIC), then asserts `Characters[CUSTOM]` is unchanged — the exact branch that was uncovered.

I also refactored the fix itself: `getPasswordConfiguration()` no longer reads the persisted value from `QtPassSettings`; `ConfigDialog` now retains the custom charset in a member set by `setPasswordConfiguration()`, so the preserve branch is self-contained and unit-testable (which is what let the new test exercise it directly).

Separately, while validating the profile-table fixes I found that the real `profileTable` sorts on a non-existent column (section 3) at construction, so `setProfiles()` corruption is not reachable on a normal open — that change is defensive hardening. The reachable table bug is `on_addButton_clicked()` after the user sorts by name; added `addProfileSelectsNewRowAfterSort` as a real regression guard (verified it fails without the pointer fix).

annejan:
Fixed in 7a7073f99 — moved the profile restore into an RAII scope guard so it runs even if an assertion fails early. Good catch on the cleanup leak (tst_configdialog has no cleanupTestCase).

PR #1604, opened 2026-07-03
fix: correct ImitatePass git re-encryption recipients, CWD, and broken copy

annejan:
## Summary

Batch 3 of the full `src/` review — three defects in the `ImitatePass` git/gpg re-encryption path (native `gpg2`/`git` mode, no `pass` CLI).

### 1. Re-encryption to the wrong recipients (`verifyGpgIdForDir`)
On a `.gpg-id` **cache hit** the function returned early **without recomputing `gpgId`**, leaving it set to the previously processed directory's recipient list. When `reencryptPath()` walks into a directory whose `.gpg-id` was already verified, files there could be re-encrypted to the **wrong keys**. Now the signature is verified once per `.gpg-id`, but the recipient list is always refreshed for the current file.

### 2. git commands run in the wrong directory (`createBackupCommit`, `reencryptSingleFile`)
Both used `Executor::executeBlocking()`, which sets **no working directory**, so `git status` / `git add` / `git commit` ran in QtPass's **launch directory** rather than the store. Depending on where QtPass was started this either aborted all re-encryption (`fatal: not a git repository`) or operated on an **unrelated repository**. Fixed with `git -C <store>`.

### 3. Broken git-mode copy (`Copy`)
Git mode invoked `git cp`, which is **not a git subcommand**, so drag-and-drop copy silently produced nothing (and the follow-up re-encryption ran against a destination that never existed). Copy now does a filesystem copy (synchronous, so the destination exists before re-encryption) and stages the new path with `git add`.

## Test

`tst_integration::imitatePass_gitCopyAndShow` sets up a real git store, copies an entry in git mode, and verifies the destination exists and decrypts to the original content. Confirmed to fail without the copy fix (`QFile::exists(dst)` is false — `git cp` created nothing).

Full suite: 22 integration tests pass, all unit suites green, zero doxygen warnings.

## Deferred to follow-ups
- **Move** does not re-encrypt for the destination folder's `.gpg-id` (unlike `pass mv` / `Copy`) — the git path uses async `git mv`, so re-encryption needs completion ordering.
- The `actualKeys != gpgId` comparison compares 16-char key IDs against raw `.gpg-id` entries (often fingerprints/emails), so the "skip already-correct files" optimization never matches — a performance issue, not a correctness one.

Related: #1598 (keygen error handling).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `.gpg-id` verification caching to ensure signature checks aren’t incorrectly reused across directories, while recipients continue to refresh as expected.
  * Improved Git-backed operations so staging/committing (re-encryption, backup commits) execute in the password-store repo and use porcelain status checks to determine when to commit.
  * Updated Git-mode copy to always write the encrypted result to disk first, then stage and commit it.

* **Tests**
  * Added a regression integration test covering Git-mode copy and confirming the copied file is tracked and can be shown/decrypted.

### DIFF
--- a/src/imitatepass.cpp
+++ b/src/imitatepass.cpp
@@ -920,14 +920,19 @@ void ImitatePass::Copy(const QString src, const QString dest,
                        const bool force) {
   QFileInfo destFileInfo(dest);
   transactionHelper trans(this, PASS_COPY);
-  QDir qDir;
   if (force) {
-    qDir.remove(dest);
+    QFile::remove(dest);
   }
   // git has no "cp" subcommand, so copy on the filesystem in both modes and,
   // when using git, stage the new path afterwards. QFile::copy is synchronous,
-  // so the destination exists before the re-encryption below runs.
-  QFile::copy(src, dest);
+  // so the destination exists before the re-encryption below runs. It fails
+  // (without overwriting) when dest already exists, so surface that instead of
+  // committing a copy that never happened.
+  if (!QFile::copy(src, dest)) {
+    emit critical(tr("Copy failed"),
+                  tr("Could not copy %1 to %2.").arg(src, dest));
+    return;
+  }
   if (m_settings.useGit) {
     executeGit(GIT_COPY, {"add", pgit(dest)});
     QString message = QString("Copied from %1 to %2 using QtPass.");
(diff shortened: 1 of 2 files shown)


annejan:
Fixed in 19c691ce2 — Copy now checks `QFile::copy`'s result and emits a critical error (aborting before the misleading git add/commit) when the destination already exists or the copy fails, and uses `QFile::remove` instead of the CWD-relative `QDir::remove`. Both good catches.

annejan:
Both addressed in 447c0a88e:
- Added a `git ls-files` assertion so the test also verifies the copy is staged/committed into the repo, not just written to disk (checking tracked state rather than the async "Copied" commit subject, since re-encryption commits synchronously first).
- Extracted `initGitRepo()` and reused it in both `imitatePass_gitCopyAndShow` and `imitatePass_gitInitAndCommit` to drop the duplicated bootstrap.

PR #1605, opened 2026-07-03
fix: prevent PasswordDialog content duplication and premature-save data loss

annejan:
## Summary

Two data-integrity bugs in `PasswordDialog` when editing an existing entry (from the full `src/` review).

### 1. Free-text body duplicated on save
`setPassword()` filled the body with `insertPlainText()`, which **appends**. `PasswordDialog` connects to the shared `Pass::finishedShow` signal, so if more than one result arrives (e.g. a stale queued `Show`), the body got a second copy appended and `getPassword()` wrote the **duplicated** content back, corrupting the file.

Fix: use `setPlainText()` so every field is **replaced** (last result wins, no duplication). The dialog deliberately keeps listening — it is modal (`d.exec()`) and always edits the selected entry, and `finishedShow` carries no request identity, so disconnecting on the first signal could consume a stale queued result and never load this dialog's own entry.

### 2. Premature save wipes the entry
`on_accepted()` guarded with `if (newValue.isEmpty()) return;`, but `getPassword()` always returns at least a trailing newline, so the guard was **dead code**. Clicking **OK before the asynchronous `Show()` finished decrypting** overwrote the existing entry with the empty dialog fields — silent data loss.

Fix: track an `m_contentLoaded` flag (set when the entry's decrypted content arrives) and refuse to save an existing entry until it has loaded. New entries are unaffected.

## Testing
No `PasswordDialog` unit-test harness exists, and constructing one needs a live `Pass` singleton plus async `Show` mocking (and dialog + offscreen teardown is fragile on Qt 5.15). Verified by construction and by the related suites (ui/filecontent/model/passwordconfig) staying green; zero doxygen warnings.

## Scope
First of the review's "misc HIGH" batch. Deferred pending per-caller analysis:
- `realpass.cpp:162` — the no-force overwrite guard: the drag-drop caller passes real `.gpg` paths and already confirms overwrite in the UI, so the guard works for that path; needs checking for any caller that passes a non-`.gpg` dest.
- `qtpass.cpp:419` (clipboard autoclear after navigation) and `pass.cpp:109` (transaction bookkeeping when a command is discarded).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed saving of existing password entries so it won’t occur until decrypted details have finished loading.
  * Prevented duplicated or accumulated dialog text when repopulating the password dialog.
  * Improved handling of displayed entry contents to avoid showing or overwriting with stale/incorrect information during rapid item changes.

### DIFF
--- a/src/passworddialog.cpp
+++ b/src/passworddialog.cpp
@@ -116,11 +116,15 @@ void PasswordDialog::on_createPasswordButton_clicked() {
  * @brief PasswordDialog::on_accepted handle Ok click for QDialog
  */
 void PasswordDialog::on_accepted() {
-  QString newValue = getPassword();
-  if (newValue.isEmpty()) {
+  // For an existing entry, refuse to save until its decrypted content has
+  // loaded (Show is asynchronous). getPassword() always returns at least a
+  // newline, so the previous isEmpty() check never fired and clicking OK early
+  // overwrote the entry with the empty dialog fields.
+  if (!m_isNew && !m_contentLoaded) {
     return;
   }
 
+  QString newValue = getPassword();
   if (newValue.right(1) != "\n") {
     newValue += "\n";
   }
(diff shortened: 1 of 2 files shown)


annejan:
Good catch — resolved in 34f948782 by dropping the disconnect. You're right that `finishedShow` carries no request identity, so disconnecting on the first signal could consume a stale queued Show and then never see this dialog's own result. Since the dialog is modal (`d.exec()`) and always edits the selected entry, no user-triggered Show can race after it opens, and `setPassword()`/`setPlainText()` replace every field, so repeated results overwrite (last wins) rather than duplicate. Kept the `m_contentLoaded` save-guard and the setPlainText fix, which are the actual data-integrity fixes.

PR #1607, opened 2026-07-03
fix: keep clipboard autoclear tracking when navigating between entries

annejan:
## Summary

A HIGH-severity clipboard-security bug from the full `src/` review: a copied password could stay on the clipboard **past its autoclear window** if the user selected another entry before it cleared.

## The bug

`clippedText` records what `copyTextToClipboard()` put on the clipboard, so the autoclear timer can verify and clear it:

```cpp
void QtPass::clearClipboard() {
  ...
  if (this->clippedText == clipboard->text(QClipboard::Clipboard)) {
    clipboard->clear(QClipboard::Clipboard);
    ...
  }
}
```

But two navigation paths overwrote that tracker before the timer fired:
- `setClippedText()` is called on **every** entry show and did `clippedText = password;`
- `clearClippedText()` was called on **every** tree click and did `clippedText = "";`

So: copy password A (timer armed) → click another entry → `clippedText` becomes B's password (or empty) while the clipboard still holds A → the timer sees `clippedText != clipboard` → **does not clear** → A stays on the clipboard indefinitely.

## The fix

The staged value was never consumed anywhere — on-demand copy uses the field's own value via `copyTextToClipboard()` — so `setClippedText`/`clearClippedText` only corrupted the autoclear tracker. Make `clippedText` a pure record of the current clipboard contents:
- `setClippedText()` now only copies in "always copy" mode (through `copyTextToClipboard()`, which sets the tracker) and no longer overwrites it otherwise.
- The vestigial `clearClippedText()` and its per-navigation call are removed.

Now selecting another entry no longer disarms a pending autoclear (always-copy mode simply re-copies and re-arms for the new entry).

## Testing
No `QtPass`-level clipboard test harness exists (needs a QApplication clipboard + timer). Verified by construction and by mainwindow/ui suites staying green; zero doxygen warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard handling: passwords are now only tracked when they are actually copied to the system clipboard, reducing unintended clipboard-related state changes in non-“always copy” modes.
  * Fixed password selection behavior: switching between entries no longer clears any previously stored “clipped text” unexpectedly.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -574,8 +574,6 @@ void MainWindow::on_treeView_clicked(const QModelIndex &index) {
   bool cleared = ui->treeView->currentIndex().flags() == Qt::NoItemFlags;
   m_currentDir = Util::getDir(ui->treeView->currentIndex(), false, model,
                               proxyModel, QtPassSettings::getPassStore());
-  // Clear any previously cached clipped text before showing new password
-  m_qtPass->clearClippedText();
   QString file = getFile(index, true);
   ui->passwordName->setText(file);
   if (!file.isEmpty() && !cleared) {
--- a/src/qtpass.cpp
+++ b/src/qtpass.cpp
@@ -415,17 +415,16 @@ void QtPass::doGitPush() {
  */
 void QtPass::setClippedText(const QString &password, const QString &p_output) {
   const AppSettings s = QtPassSettings::load();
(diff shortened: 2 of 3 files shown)


annejan:
Updated in de6e4b5e8 — the docstring now states it only copies (and updates the tracker) in "always copy" mode, and is a no-op otherwise.

PR #1608, opened 2026-07-03
test: add unit tests for PassBackendFactory backend selection

annejan:
## Summary

`PassBackendFactory` is the core router that decides whether QtPass drives the real `pass` CLI (`RealPass`) or gpg/git directly (`ImitatePass`), caches that choice, and exposes `invalidate()` so a settings change forces a rebuild. It had **no dedicated tests** — this adds a `tst_passbackendfactory` binary.

## Coverage (no gpg/pass needed — pure routing/lifecycle)

- `usePass=false` selects `ImitatePass`; `usePass=true` selects `RealPass`
- `getPass()` caches the instance across calls (same pointer)
- switching `usePass` rebuilds the backend for the new mode — documents that `setUsePass()` calls `PassBackendFactory::invalidate()` (the same invalidation the ConfigDialog profile flow relies on)
- `getPass()` creates the store directory when it does not yet exist

All assertions check the returned backend **type** (`dynamic_cast`) and lifecycle, so the tests run on plain offscreen CI without a keyring.

Runs green locally alongside the full suite; settings are saved/restored around the run.

## Summary by CodeRabbit

* **Tests**
  * Expanded automated test coverage for password and key management flows, including backend selection, configuration changes, and directory creation behavior.
  * Added new test setup to better validate app behavior across supported environments.

### DIFF
--- a/tests/auto/auto.pro
+++ b/tests/auto/auto.pro
@@ -1,5 +1,5 @@
 TEMPLATE = subdirs
-SUBDIRS += util ui model settings passwordconfig filecontent simpletransaction gpgkeystate exportpublickeydialog importkeydialog keygendialog trayicon configdialog locale mainwindow userinfo profileinit grepsearchcontroller passworddisplaypanel
+SUBDIRS += util ui model settings passwordconfig filecontent simpletransaction gpgkeystate exportpublickeydialog importkeydialog keygendialog trayicon configdialog locale mainwindow userinfo profileinit grepsearchcontroller passworddisplaypanel passbackendfactory
 win32: SUBDIRS -= executor
 !win32: SUBDIRS += executor
 !win32: SUBDIRS += integration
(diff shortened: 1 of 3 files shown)


annejan:
Thanks, but I don't think this one holds — it conflates two different methods:

- `getPass()` calls `pass->init(s)` — that's `Pass::init(const AppSettings&)` (pass.cpp:123), which only stores the settings snapshot and adjusts the process `env` (PATH/gpgHome). No process is launched.
- `RealPass::Init(path, users)` (realpass.cpp:96, capital I) is the store-initialization operation that runs `pass init` — but nothing in these tests calls it.

So constructing a RealPass and calling `init(s)` spawns nothing, and the assertions only check the returned backend **type** (`dynamic_cast<RealPass*>`), which holds regardless of whether `pass` is installed or working. The suite genuinely has no runtime dependency on the `pass`/gpg binaries (it passed on all CI platforms). Resolving as a false positive.

PR #1611, opened 2026-07-04
test: GUI end-to-end — selecting a tree entry decrypts into the panel

annejan:
## Summary

The **first test that drives the real QtPass GUI** — step 2 of the e2e effort. Everything so far tested backends or constructed widgets in isolation; this one simulates an actual user action through the full stack.

## What it does
1. Inserts a real entry (`guitest` with a `url` field) into a temp store via real gpg.
2. Constructs `MainWindow` over that store (gpg mode, gpg path pre-set so no config dialog).
3. Waits for the asynchronous `QFileSystemModel` to expose the entry in the tree, then captures its index.
4. Clicks it with `QTest::mouseClick` on the tree viewport.
5. Polls (`QTRY_VERIFY_WITH_TIMEOUT`) until the display panel's `QTextBrowser` renders the decrypted field content.

So it exercises the full **tree-click → `Show` → gpg decrypt → panel-render** path.

## Reliability
The two async gaps (model load, decryption) are absorbed by `QTRY_VERIFY_WITH_TIMEOUT`. Verified stable over repeated local runs. Uses the offscreen platform like the other widget tests.

This is the non-modal user flow; modal Add/Edit dialogs (driven via a `QTimer` poker) would be a follow-up if this proves stable on CI.

Integration suite: 26 tests pass.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end GUI integration coverage for selecting an entry and verifying its decrypted content is displayed correctly.
  * Improved reliability of the integration test by handling asynchronous loading and polling the UI until the expected decrypted value appears.

### DIFF
--- a/tests/auto/integration/tst_integration.cpp
+++ b/tests/auto/integration/tst_integration.cpp
@@ -1016,13 +1016,27 @@ void tst_integration::mainWindow_selectingEntryShowsDecryptedContent() {
     QVERIFY2(waitForSignal(insertSpy), gpgInsertErrorMsg(errSpy));
   }
 
+  // Restore every globally-mutated setting on any exit path (including early
+  // QVERIFY returns): this test changes passStore/usePass/gpgExecutable/
+  // hideContent, all of which live in AppSettings.
+  const AppSettings savedSettings = QtPassSettings::load();
+  struct SettingsRestorer {
+    AppSettings s;
+    ~SettingsRestorer() {
+      QtPassSettings::save(s);
+      PassBackendFactory::invalidate();
+    }
+  } restorer{savedSettings};
+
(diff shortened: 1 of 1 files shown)


annejan:
Both addressed in db5922095:
- Added a `SettingsRestorer` scope guard that snapshots `AppSettings` up front and restores it (plus `PassBackendFactory::invalidate()`) on any exit path — so passStore/usePass/gpgExecutable/hideContent are all restored. Also set `hideContent = false` explicitly, since `passShowHandler()` gates the panel on hideContent (not hidePassword).
- Added `QVERIFY2(QTest::qWaitForWindowExposed(&w), ...)` after `w.show()` before any visualRect/click.

PR #1661, opened 2026-09-13
fix: follow light/dark theme switches at runtime

annejan:
## Problem

KDE switches the palette between light and dark in the morning and evening (GNOME and macOS do the same). After a switch, parts of the QtPass main window keep the previous theme:

In the report (light theme active) the content area under the password field was still the dark theme's black, and the field frame kept its dark border.

## Cause

Two places bake palette colours into Qt stylesheets, and `QStyleSheetStyle` resolves them **once at polish time**; a runtime palette change does not repolish.

1. **`ui->textBrowser`** had `styleSheet = "background: palette(base)"` in `mainwindow.ui`. That is what a `QTextBrowser` paints by default anyway, so the stylesheet bought nothing and cost the live palette. It was also the reason the context-menu reparent workaround in `showBrowserContextMenu()` exists.
2. **`PasswordDisplayPanel::createFieldFrame()`** formats `palette().color(QPalette::Mid).name()` into the field-frame border stylesheet, so an entry that is already displayed keeps the old border colour until you select another one.

## Fix

- Remove the `textBrowser` stylesheet (`mainwindow.ui`). Native Base painting follows the palette. Context-menu workaround kept, comment updated to explain why it stays.
- `PasswordDisplayPanel`: border style moved into `applyFrameStyle(QFrame*)`; new `refreshPalette()` re-applies it to every frame in the grid.
- `MainWindow::changeEvent()`: on `QEvent::PaletteChange` call `m_displayPanel->refreshPalette()`. Note: for an application-wide switch, top-level *widgets* receive `PaletteChange`; `ApplicationPaletteChange` is delivered to the `QApplication` object (verified with a probe — first attempt used the wrong one).

No refresh-on-show hook needed: with the stylesheet gone Qt repaints the browser itself, and the frame refresh is event-driven.

## Tests (`tst_mainwindow`, 19/19)

- `textBrowserFollowsRuntimePaletteChange` — shows the window, sets Base to `#101010`, grabs the browser and checks the centre pixel; then `#fafafa`, same check. **Red on the old code**: `Actual #ffffff, Expected #101010` — the browser is frozen at its polish-time colour.
- `fieldFrameBorderFollowsRuntimePaletteChange` — displays an entry, sets `QPalette::Mid` to `#123456`, asserts the frame stylesheet now contains it.

`tst_passworddisplaypanel` 21/21, clang-format clean, doxygen clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved light and dark theme switching so password display frames update their border styling immediately.
  - Ensured the content browser background follows runtime palette changes.
  - Removed conflicting fixed styling so theme-dependent colors render correctly.
  - Updated existing password fields and browser content when the application palette changes.

- **Tests**
  - Added coverage for runtime palette changes affecting content backgrounds and password field borders.


annejan:
Qt 5.15 job: offscreen Fusion tints Base `#101010` to `#070c10`, so the exact pixel `QCOMPARE` failed there. Switched to lightness thresholds (`< 64` after dark palette, `> 192` after light). Still red on the old stylesheet (`browser still light after dark palette: #ffffff`), green with the fix, 19/19 locally.

annejan:
Qt 5.15 offscreen again: this run painted `#71c1f5` after the light palette (last run `#070c10` after dark) — pixel grabs there are not deterministic. Test now asserts `textBrowser->styleSheet().isEmpty()` on every Qt (red on the old `.ui`: `textBrowser carries a stylesheet: background: palette(base)`), and keeps the palette-swap pixel check for Qt 6 where it reproduced the report. `QSKIP` on Qt 5.

PR #1669, opened 2026-09-13
fix: toolbar stays in the old theme after a KDE light/dark switch

annejan:
Follow-up to #1661. With that merged, the content pane and field frames follow a runtime theme switch — the **toolbar** on KDE still did not: dark bar with (correctly recoloured, therefore dark) icons on a light window.

## Diagnosis (instrumented build on Plasma 6 / Breeze, `plasma-apply-colorscheme` flips)

| after flip to light | value |
|---|---|
| `QApplication::palette()` Base | `#ffffff` ✓ |
| `textBrowser` pixel | `#ffffff` ✓ (#1661) |
| toolbar icon average | `#232629` ✓ (recoloured for light) |
| **`ui->toolBar` palette Window** | **`#292c30`** ✗ still dark, 4 s later too |

The toolbar carries `breeze_has_toolsarea_palette` and `WA_SetPalette`: Breeze's `ToolsAreaManager` gives top toolbars a *header* palette and re-applies it on every `ApplicationPaletteChange` and on re-polish ([`breezetoolsareamanager.cpp`](https://invent.kde.org/plasma/breeze/-/blob/master/kstyle/breezetoolsareamanager.cpp), `configUpdated()`). It derives the header colours from `KSharedConfig::openConfig(kdeglobals)`, which in this process is not reparsed, so it keeps stamping the previous theme's colours back on. `style()->unpolish()/polish()` on the toolbar does not help (same stale manager palette). A minimal Qt probe app happened to get fresh colours; QtPass doesn't — I couldn't pin down the difference inside KConfig, and it isn't ours to fix anyway.

## Fix

`MainWindow` filters `PaletteChange` on `ui->toolBar`; a zero-timer then compares the lightness of the toolbar's `Window` colour with the application palette's. A genuine header tint differs by a few units; the other theme by >64. In that case the imposed palette is cleared so the toolbar follows the application palette. Because it hooks the toolbar's own `PaletteChange`, any later stale re-apply by the style is caught the same way. Verified live: light flip → toolbar `#eff0f1` and stays; dark flip → `#292c30`.

## Tests (`tst_mainwindow`, 21/21)

- `toolBarDropsStaleStylePaletteAfterThemeSwitch` — sets the Breeze property and a dark palette on the toolbar while the app palette is light; expects `WA_SetPalette` cleared and the window colour to match. **Red on `main`.**
- `toolBarKeepsHeaderTintInSameTheme` — a same-theme tint (`#e3e5e7` on `#eff0f1`) is left alone.

clang-format and doxygen clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed toolbar colors not updating correctly when switching between light and dark themes.
  * Preserved intentional toolbar header tinting when it matches the active theme.

* **Tests**
  * Added coverage for toolbar palette updates during theme changes.
  * Added coverage ensuring valid same-theme toolbar styling remains intact.


annejan:
Second commit: clearing the palette alone was not enough on screen — Breeze *also* paints the main window's tools-area rectangle from the same stale palette, and `QToolBar` is transparent by default, so the first version only looked right in `QWidget::grab()`. Now `setAutoFillBackground(true)` accompanies the reset. Verified with an on-screen pixel grab (`#fcfcfc` after the light flip, `#292c30` after dark) and a screenshot of the live window: light theme, light bar, dark icons. Test asserts the autofill too. 21/21.

Issue #1682, opened 2026-09-13
QtPass 2.0 backlog (2.0-nazomer)


annejan:
Tracking issue for the 2.0 series (Qt 5 dropped in #1679; umbrella #908).

**Where the work happens (since 2026-09-16):** `main` is 2.0 development (#1746 merged the
`2.0-nazomer` branch into it; that branch is deleted). 1.8.x fixes go to the `1.8`
maintenance branch, which shipped [v1.8.1](https://github.com/IJHack/QtPass/releases/tag/v1.8.1).
Open PRs for items below against `main`.

**How this list was made.** An AI-assisted review (Claude Code) swept `src/`, `main/`, `tests/` and the build files from eight angles — security, correctness/architecture, Qt 6 idioms, C++ cleanup, dead code, build & packaging, tests, UX. Every raw finding was then checked by two independent verifier passes against the actual code on `2.0-nazomer`; only findings that survived are listed (file:line cited for each), the rest are under "Refuted" at the bottom so nobody re-litigates them. Effort tags and the "not doing" section are a proposal — argue in the comments, tick as things land, split anything M/L into its own issue when picked up.

108 verified findings, merged into 84 items. Effort: **S** ≈ hours, **M** ≈ a day or two, **L** ≈ a week+. **breaking** = changes user-visible behaviour or stored data.

## Top 10 for 2.0

- [x] **`addFolder` writes a zero-byte `.gpg-id`** — `src/mainwindow.cpp:1712` · S · breaking — `addGPGId` defaults on, `user.enabled` is never set outside UsersDialog, so the loop writes nothing; the empty file shadows the parent's recipients and every Insert in that folder fails. Shipped broken since #742. ✅ #1688, #1695
- [x] **WSL launches go through the login shell, not `--exec`** — `src/executor.cpp:34` · S — entry paths, `.gpg-id` recipients and commit messages are word-split and `$()`-expanded by the WSL shell; a hostile shared store gets code execution. ✅ #1686
- [x] **Re-encryption blocks the GUI thread and pops one modal per failed file** — `src/imitatepass.cpp:700` · M — runs after every UsersDialog OK on the default backend; 3 blocking gpg processes per entry, no progress, no cancel, dead watchdog. ✅ #1697
[...]

PR #1691, opened 2026-09-14
fix: surface unset executables instead of wedging the command queue (#1682)

annejan:
## Summary
- Does not silently drop a command whose executable is unset (e.g. `useGit` enabled with no git binary configured). Previously the Executor dropped it, the transaction never completed, and its stale head swallowed every later completion signal for the rest of the session.
- `Executor::executeNext()` now fails an empty executable through the same deferred `error()` path as a failed-to-start process, so the queue keeps draining.
- `ImitatePass::executeGit()` skips git commands outright when `gitExecutable` is unset, so the parent operation (gpg/filesystem) still completes cleanly.
- Regression tests: executor unit test (empty executable emits error + subsequent command still runs) and integration test (insert completes when git is unconfigured, and a following operation still completes).


annejan:
Adversarial review pass

- `src/imitatepass.cpp:963` — the new `executeGit()` early return turns `Remove()` and `Move()` into silent no-ops when `useGit` is on but `gitExecutable` is empty (the #1682 configuration): both take the `useGit` branch, `executeGit(GIT_RM/GIT_MOVE)` and `gitCommit()` return immediately, the `else` filesystem fallback (`QFile::remove` / `removeRecursively` / `QDir::rename`) never runs, `transactionEnd()` sees `lastInTransaction == INVALID` so nothing is queued, and no `finishedRemove`/`finishedMove`/`processErrorExit` ever fires — the `.gpg` file stays on disk, the tree is not refreshed, and no error or status text is shown. Less severely, `GitPull`/`GitPush` from the toolbar show "Updating password-store" and do nothing, and `Insert`/`Copy` encrypt but never stage/commit, so the store silently drifts from git. The comment's "the parent operation (gpg/filesystem) still completes cleanly" does not hold for Remove/Move, and this inverts the PR title: before the guard, the Executor would at least have reported "No executable configured for this command". Suggested fix: either surface it once from `executeGit()` via a queued `QMetaObject::invokeMethod` emitting `processErrorExit(1, tr("No git executable configured"))` (as `Pass::GenerateGPGKeys` does), or compute `const bool gitAvailable = m_settings.useGit && !m_settings.gitExecutable.isEmpty();` in `Remove()`/`Move()`/`Insert()`/`Copy()`/`GitPush()` so the filesystem fallback runs and a finished signal is produced, plus `emit statusMsg(tr("Git executable not configured; skipping git"))`. Add an integration test for `Remove` with `useGit=true` and empty `gitExecutable` asserting the file is gone and `finishedRemove` fires.

Automated review (Claude Code) on behalf of the maintainer; findings verified against the branch.

annejan:
Self-review pass — applied the finding above.

- `Remove()`, `Move()`, `Insert()`, `Copy()`, `Init()`, `GitPull()`, `GitPush()` now branch on `gitReady()` (git enabled **and** executable set; emits `statusMsg("Git executable not configured, skipping git")` once per operation). The re-encryption worker's per-file git and auto pull/push use the message-free `gitConfigured()`.
- `executeGit()` no longer returns early on an empty executable; with the Executor change in this PR an empty executable that still gets there surfaces as "No executable configured for this command" instead of vanishing.
- Tests: `imitatePass_removeFallsBackWhenGitNotConfigured` and `imitatePass_moveFallsBackWhenGitNotConfigured` — file removed / renamed on disk, no `processErrorExit`, status message present, a following `Insert()` completes. Both fail on the previous commit (`'!QFile::exists(gpgFile)' returned FALSE`), pass now. Full `make check`: 26 suites, 933 passed, 0 failed.
- Branch rebased onto current `2.0-nazomer` (includes #1697's worker), force-pushed with lease.

PR #1693, opened 2026-09-14
Fix open-in-browser tooltip showing &amp; in URLs (#1682)

annejan:
## Summary

- The password pane's open-in-browser button showed `Open https://example.org/?a=1&amp;b=2 in browser` on hover for any URL with a query string.
- Force the tooltip through Qt's rich-text path so the HTML escaping is decoded again.
- The translatable string `Open %1 in browser` is unchanged (64 locales untouched); the URL handed to the browser is still the raw validated value.
- New data-driven regression test in `tst_passworddisplaypanel`.

## Root cause

`PasswordDisplayPanel::addField()` HTML-escapes the URL and calls `setToolTip()`. `QToolTip` renders with `Qt::AutoText`, so the text is only treated as HTML when `Qt::mightBeRichText()` returns true, which needs a `<tag>` or a literal `&lt;`. An escaped `&amp;` alone does not qualify, so the entity was drawn literally. Same root cause as #1683, same function, still unfixed for the tooltip.

## Fix

Wrap the tooltip in `<qt>…</qt>` (`qt` is in `QTextHtmlParser`'s element table, so `mightBeRichText()` is always true). Dropping the escaping and using plain text was considered and rejected: a URL may legitimately contain `&lt;` in its query (`&`, `l`, `t`, `;` are all valid URL characters), and auto-detection would then render it as HTML, so escaping plus a forced rich-text marker is the only variant that is right for every launchable URL.

## Tests

Added `urlButtonToolTipShowsUrlVerbatim` (data-driven) to `tests/auto/passworddisplaypanel/tst_passworddisplaypanel.cpp`. It locates the open-in-browser button, renders its tooltip the way `QToolTip` does (auto-detect, then `QTextDocument`) and asserts the URL appears verbatim without `&amp;`, for a plain URL, a query string and an entity-looking query (`?q=&lt;x&gt;`).

- Before the fix: 2 of 3 rows fail (`Open https://example.org/?a=1&amp;b=2 in browser`, `...?q=&amp;lt;x&amp;gt;...`).
- After the fix: `tst_passworddisplaypanel` 33 passed; `tst_mainwindow` 21 passed; `tst_ui` 55 passed (`QT_QPA_PLATFORM=offscreen`, Qt 6.11.2).

## Notes

- Behaviour change: the tooltip is now rich text, so it word-wraps long URLs instead of running as one line; no other visible difference.
- No header changes; only `src/passworddisplaypanel.cpp` and the test suite touched.

Part of #1682.


annejan:
## Self-review pass

**Fixed** (125765fbe)

- **Tooltip word-wrapped into a cramped box.** Confirmed on Qt 6.11.2 against the real `QToolTip` widget (offscreen): the bare `<qt>` wrapper makes `QTipLabel` enable word wrap, and `QLabel`'s wrapped-size heuristic then folds the tooltip into 3-4 lines, breaking the URL at `/` and `?` (`?a=1&b=2` -> 114x51 instead of the base branch's 298x17; the plain `/login` URL without any `&` folded too). `<nobr>` does not help (Qt only turns its spaces into non-breaking ones). The wrapper is now `<qt style="white-space:nowrap">`, which keeps the rich-text path (entities decoded) but lays every tooltip out on one line again: 269x17 / 283x17 / 848x17 for the three test URLs.
- **Test coverage.** `urlButtonToolTipShowsUrlVerbatim` now also asserts single-line layout (the `QLabel` size hint with word wrap on, as `QTipLabel` sets it, must equal the size hint with word wrap off) and gained a long sign-in URL row. The assertion fails on all four rows with the old wrapper and passes with the fix.
- `tst_passworddisplaypanel` 34 passed, `tst_mainwindow` 21 passed, `tst_ui` 55 passed (`QT_QPA_PLATFORM=offscreen`).

**Notes**

- The "Behaviour change: the tooltip is now rich text, so it word-wraps long URLs" line in the description is obsolete; the tooltip is single-line again. Remaining difference from the base branch: a URL wider than the whole screen no longer gets folded to fit (the plain-text tooltip did that as a fallback) and is clipped at the screen edge instead.
- CodeRabbit skipped this PR (non-default base branch), so no bot findings to address.

PR #1697, opened 2026-09-14
Run re-encryption on a worker thread with progress and cancel (#1682)

annejan:
## Summary
- `ImitatePass::reencryptPath()` now runs its scan and per-file gpg/git work on a `QThread::create` worker (same pattern as `Grep()`), so the GUI no longer freezes after a UsersDialog OK, a Move or a Copy on the native backend.
- New `reencryptProgress(current, total)` signal and `cancelReencryptPath()`; MainWindow shows a cancellable `QProgressDialog`. Cancel finishes (and commits) the file in progress, then stops; the status bar reports how many files were checked and re-encrypted.
- Per-file failures are collected and reported in one `critical()` at the end (capped at 15 listed names plus "... and N more") instead of one modal per file from inside the loop.
- The public `startReencryptPath` / `endReencryptPath` / `critical` / `statusMsg` contract used by `QtPass`/`MainWindow` is unchanged; `GitPush()` and everything touching the Executor queue stays on the owning thread.

## Root cause
`reencryptPath()` was fully synchronous: for every entry it ran `gpg --list-only`, `gpg -d`, `gpg -e`, a verification `gpg -d` and a `git commit`, all through `executeBlocking`, on the GUI thread. The event loop was blocked, so there was no repaint, no progress, no cancel, and the UI watchdog could not fire. `emit critical(...)` sat inside the loop, so a store with N undecryptable files produced N modal dialogs.

## Fix
- `reencryptPath()` sets an active flag, emits `startReencryptPath()` synchronously and calls `startReencryptWorker()`. A second call while a run is active is ignored (status message).
- `startReencryptWorker()` waits (100 ms poll) until the `Executor` queue is idle — new `Executor::isIdle()` — because `Init()`, `Move()` and `Copy()` queue `git add`/`git commit` right before calling `reencryptPath()`, and the worker's blocking git calls would otherwise compete with them for the index lock. Then it starts the worker.
- `reencryptFiles()` (worker thread) does the optional pull, the backup commit, the `.gpg-id` verification per directory and the per-file re-encryption, emitting `reencryptProgress` per file and checking the atomic cancel flag between files. Signals emitted from the worker are delivered queued to their GUI-thread receivers.
- `finishReencrypt()` (owning thread, via queued `invokeMethod`) emits the aggregated `critical`, the summary `statusMsg`, pushes when configured (not after a cancel or abort, so the partial result can be inspected first) and emits `endReencryptPath()`.
- The destructor cancels and joins the worker (it calls member functions, unlike the grep workers).
- `MainWindow`: `startReencryptPath()` shows the progress dialog and stops the watchdog; `reencryptProgress()` updates it; `endReencryptPath()` closes it. Because the run is now asynchronous, the `finishedInit`/`finishedMove`/`finishedCopy` completions (which call `setUiElementsEnabled(true)`) arrive while files are still being rewritten, so `setUiElementsEnabled(true)` is ignored while a re-encryption is running.

## Tests
- New suite `tests/auto/imitatepass/` (registered in `tests/auto/auto.pro`), no gpg needed — the gpg path is unstartable or a slow failing shell script:
  - `reencryptPathEmitsStartSynchronouslyAndEndLater` — `startReencryptPath` fires before `reencryptPath()` returns, `endReencryptPath` only afterwards (fails on the old synchronous code).
  - `reencryptPathAggregatesFailuresIntoOneCritical` — 3 failing files yield exactly one `critical` listing all three, `(0,3)…(3,3)` progress, and a "3 failed" summary (old code emitted 3 criticals).
  - `reencryptPathCancelStopsBetweenFiles` (Unix) — cancel after the first progress step stops with 1 ≤ checked < 6, the aggregated dialog lists exactly the checked files, a nested `reencryptPath()` during the run does not start a second run, and the summary says "cancelled".
- `tests/auto/mainwindow`: `reencryptKeepsUiDisabledUntilEnd` — progress dialog appears once (idempotent start), `reencryptProgress(2,5)` updates it, `setUiElementsEnabled(true)` is ignored during the run, `endReencryptPath()` re-enables the UI and removes the dialog, and the normal enable/disable cycle works afterwards.
- `tests/auto/integration`: the two re-encryption tests now wait for `endReencryptPath` instead of assuming a synchronous call; both pass against real gpg.
- Ran: `qmake6 -r`, `make -j3 sub-src sub-main`, `QT_QPA_PLATFORM=offscreen make -j3 check` — all suites pass (integration: 27 passed; imitatepass suite stable over 5 repeated runs). `clang-format --dry-run --Werror` clean on all touched C++ files.

## Notes
- Behaviour change: after a **cancel** or an abort (backup-commit or `.gpg-id` verification failure) no automatic `git push` is made; the files already re-encrypted are committed locally and get pushed with the next change. A completed run pushes as before.
- The rare `critical` emitted inside `reencryptSingleFile()` when the final rename fails ("Original has been restored") is unchanged and still arrives as its own (queued) dialog.
- `m_settings` is read from the worker while the UI (and therefore the config dialog / profile switch) is disabled; `init()` is not expected to run during a re-encryption.
- New user-visible strings (`tr`) were added; `localization/*.ts` were deliberately not touched.

Part of #1682.


annejan:
## Self-review pass

**Fixed** (ba0ed9e4a)

- `MainWindow::reencryptProgress()` dereferenced `m_reencryptProgress` after `QProgressDialog::setValue()`. On a modal dialog `setValue()` calls `QCoreApplication::processEvents()` once the dialog's minimum-duration timer has fired (checked against the Qt 6.11 source), which delivers the `finishReencrypt()` the worker queues right behind its final progress report. `endReencryptPath()` then hides the dialog and nulls the `QPointer` while `reencryptProgress()` is still on the stack, and the following `setLabelText()` was a null dereference. Fix: set maximum and label first, `setValue()` last, nothing after it.
- New test `tst_mainwindow::reencryptProgressSurvivesQueuedEnd` posts `endReencryptPath()` via a queued `invokeMethod` before the final progress call, mirroring the worker's ordering. It segfaults in `QProgressDialog::setLabelText` on the previous code and passes now; 25 repeated runs, no flakes. Full `tst_mainwindow` (23) and `tst_imitatepass` (5) pass.

**Skipped**

- Nothing. The CodeRabbit review was skipped on this base branch, so there were no bot findings to verify.

PR #1706, opened 2026-09-14
Use QStandardPaths::findExecutable for PATH lookups (#1682)

annejan:
## Summary
- `Util::findBinaryInPath(binary)` keeps its signature but is now implemented on `QStandardPaths::findExecutable` through a new `findBinaryInPath(binary, searchPaths)` overload; the one-argument form still splits Util's augmented PATH (MacGPG2, `/usr/local/bin`, WinGPG/GnuPG) and keeps the Windows WSL fallback.
- The overload drops empty PATH entries, returns nothing for an empty directory list (instead of `QStandardPaths` falling back to the process PATH), and only matches regular executable files, so a directory named like the binary is skipped.
- `ConfigDialog::isQrencodeAvailable` no longer runs `which qrencode` in the constructor: a configured path that is an executable file is accepted, otherwise `Util::findBinaryInPath("qrencode")` is used and only a hit is stored.
- Four new `tst_util` cases cover the overload.

## Why
The hand-rolled walk split PATH without `SkipEmptyParts`, accepted anything `QFileInfo::isExecutable()` returned true for (a directory `…/gpg/` on a PATH entry satisfied it), and only knew about `.exe` on Windows. `QStandardPaths::findExecutable` handles regular-file-only matching, PATHEXT, and path cleanup, and is the idiomatic Qt 6 way to do this. The `which` subprocess blocked the ConfigDialog constructor for up to two seconds, searched the unaugmented process PATH (so a Homebrew-only qrencode on macOS was missed), and on failure wrote `""` over any qrencode path the user had configured.

## Tests
- Added `findBinaryInPathSkipsDirectoryNamedLikeBinary` (fails on the old algorithm: it returned the directory), `findBinaryInPathSkipsNonExecutableFile`, `findBinaryInPathEmptyEntriesDoNotResolveToCwd` (also checks that an explicit `.` entry still resolves) and `findBinaryInPathEmptySearchPathsFindsNothing`. Unix-only with `QSKIP` on Windows, like the surrounding executable tests.
- `qmake6 qtpass.pro`, `make -j3 sub-src sub-main`, `QT_QPA_PLATFORM=offscreen make -j3 check`: 26 suites, 0 failed (tst_util 151 passed, 4 skipped as before).

## Notes
- Behaviour change: `isQrencodeAvailable` now returns true when the stored `qrencodeExecutable` points at an executable file even if `qrencode` is not on PATH, and never overwrites a stored path with `""`. A successful PATH lookup still refreshes the stored path, as before.
- Correction to the item text: the old code did not resolve an empty PATH entry to the cwd; `"" + "/" + binary` gave `/<binary>` at the filesystem root. The new test pins the intended semantics (empty entries are ignored) either way, and guards against a `QDir("")`-based reimplementation, which would resolve to the cwd.
- `findBinaryInPath` was already `const QString &` and all callers (`qtpasssettings.cpp` `initExecutables`, `configdialog.cpp` autodetect, `tst_mainwindow`) are unchanged; nothing from #1683-#1698 had touched this code.
- The private `isQrencodeAvailable` gained a `configuredPath` parameter so the constructor can pass the already-loaded `AppSettings` value instead of re-reading settings.

Part of #1682.


annejan:
## Self-review pass

**Fixed** (75cacbb52)

- `tests/auto/util/tst_util.cpp` — `findBinaryInPathEmptyEntriesDoNotResolveToCwd` compared the `"."` lookup result against `cwd.path()` by plain string. `QStandardPaths::findExecutable()` resolves relative entries through `QDir::current()` (physical `getcwd()` path), so the strings differ whenever the temp dir sits behind a symlink — the default on macOS (`/var/folders` -> `/private/var`). Reproduced on Linux with `TMPDIR=<symlink>`: FAIL before, PASS after. Now compares `QFileInfo(...).canonicalFilePath()` on both sides (same approach as `tst_executor`), with a non-empty check first so `""` vs `""` cannot pass vacuously.

**Skipped**

- Nothing else to address: no inline review comments on the PR, and CodeRabbit skipped the review because the base is `2.0-nazomer`.

Validation: `make -j3 sub-src sub-main`, `make -j3 -C tests/auto/util`, `tst_util` full suite 151 passed / 0 failed under both a plain and a symlinked `TMPDIR`. Only a test file changed; no `src/` header touched.

Issue #1709, opened 2026-09-14
1.8.1: backport the security and data-loss fixes from 2.0-nazomer


annejan:
Backport plan for a 1.8.1 patch release from `main` (still Qt 5.15 + Qt 6). Candidates are the fixes merged on `2.0-nazomer` from the verified backlog (#1682). Each was dry-run cherry-picked onto current `main`; "clean" = applies without conflict, everything else names the conflicting file. Nothing here needs Qt 6-only API (`std::as_const` and `qsizetype` are fine on 5.15).

## Must — security / data exposure

- [x] **WSL commands via `wsl --exec`** — #1686 · clean · Windows-only. Entry paths, `.gpg-id` recipients and commit messages were word-split and `$()`-expanded by the WSL login shell → code execution from a hostile shared store. ✅ #1723
- [x] **Keygen passphrase kept out of the visible template** — #1687 + follow-up #1694 · conflicts in `src/keygendialog.cpp` (+ its test) against the 1.8 version of the file; hand-port. Passphrase was echoed in clear text on screen. ✅ #1718
- [x] **Backup commit stages tracked files only** — #1685 + #1698 · both clean. `git add -A` committed and auto-pushed any stray file in the store (plaintext export, swap file) to the shared remote. #1698 is required with it (stages the new-folder `.gpg-id` that `add -A` used to catch). ✅ #1713
- [x] **`.gpg-id` recipients gpg accepts are no longer dropped, then erased** — #1684 · clean. v6 fingerprints, user IDs, `=exact` selectors were silently skipped on encrypt and removed from `.gpg-id` on the next UsersDialog save. ✅ #1710

## Should — data loss / wrong secret shown

- [x] **New folder seeded from parent recipients, never unsigned** — #1688 + #1695 · both clean. `addFolder` wrote a zero-byte `.gpg-id`; every insert in that folder failed (broken since #742). ✅ #1714
- [x] **PasswordDialog inert until the decrypt lands** — #1690 · clean. Edits typed before the content arrived were discarded or overwritten. ✅ #1722
- [x] **Password pane `&amp;` / `&quot;` / `&gt;` rendering** — #1683 + #1693 · both clean. Users read (and typed) the wrong secret for roughly a third of generated passwords. ✅ #1712

## Maybe — behaviour fixes, slightly more surface

[...]

PR #1712, opened 2026-09-14
Backport: password pane and URL tooltip HTML entity fixes (#1709)

annejan:
## Summary

Backport of the two 1.8.1 HTML-entity fixes from 2.0-nazomer to main:

- The password pane no longer shows `&amp;`, `&quot;` and `&gt;` for `&`, `"` and `>` in field values and in the unhidden password row.
- The "Open %1 in browser" tooltip decodes the escaped URL again and stays on one line.

## Original fix

`PasswordDisplayPanel::addField()` HTML-escapes every value so URLs can be wrapped in `<a>` tags and then hands the result to `QTextBrowser::setText()`. `setText()` only takes the rich-text path when `Qt::mightBeRichText()` sees a `<`, so a value without a URL and without `<` stayed plain text with the escaping still applied (`a&b` read `a&amp;b`). The URL button tooltip had the same root cause: the escaped URL went to `setToolTip()`, and QToolTip's auto-detection left `?a=1&amp;b=2` literal.

## Fix

Cherry-picked as-is (both applied cleanly, no hand-porting needed):

- 9408db144 (#1683): decide the text mode explicitly. No URL matches: `setPlainText(trimmedValue)`. URL matches: build the escaped, anchor-wrapped string and call `setHtml()`.
- 1eb6f2cf0 (#1693): wrap the tooltip in `<qt style="white-space:nowrap">…</qt>` so it always renders as rich text (decoding the entities) and does not word-wrap. Plain text is not a safe alternative because a URL may legitimately contain `&lt;`. The launched URL is still the unescaped validated value; the translatable string is untouched.

No API newer than Qt 5.15 is used (`Qt::mightBeRichText`, `QTextDocument`, `QLabel::sizeHint`, `qsizetype` already in main).

## Tests

Added in `tests/auto/passworddisplaypanel/tst_passworddisplaypanel.cpp` (came with the cherry-picks, unchanged):

- `fieldValueRendersHtmlSpecialsVerbatim` (data-driven: `&`, `"`, `>`, `<`, `'`, mixed, entity-looking text)
- `visiblePasswordRendersHtmlSpecialsVerbatim`
- `fieldValueWithUrlStillLinksAndEscapes`
- `urlButtonToolTipShowsUrlVerbatim` (data-driven: plain, query string, entity-looking query, long sign-in URL; also asserts single-line layout)

Ran locally with Qt 6: `qmake6 && make -j2 sub-src sub-main`, then `QT_QPA_PLATFORM=offscreen make -j2 check`: 24 suites, 889 passed, 0 failed. `clang-format --dry-run -Werror` clean.

## Notes

- Behaviour change: values without a URL are now shown verbatim via `setPlainText()`; values with a URL still go through the rich-text path with anchors.
- Nothing dropped or adapted; both commits were clean cherry-picks with the `(cherry picked from commit …)` lines kept.

Backport of #1683, #1693 for #1709.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed display of special characters in password and field values so they appear as entered.
  * Preserved clickable links while correctly escaping URL text.
  * Updated URL tooltips to show the full address on a single line without encoded characters.


annejan:
Before/after render (real widgets, offscreen): main vs #1712

PR #1719, opened 2026-09-14
Only link launchable http(s) URLs in inline anchors (#1682)

annejan:
## Summary

Inline link anchors in the password display panel and the main text browser were built from every `Util::protocolRegex()` match, including `ssh://`, `sftp://`, `ftp://`, `webdav://` and URLs with `user:pass@` credentials. Both `QTextBrowser`s call `setOpenExternalLinks(true)`, so clicking such an anchor handed the URL straight to the OS URL handler, bypassing `Util::isLaunchableWebUrl()`.

## Root cause

`src/passworddisplaypanel.cpp` (field values) and `src/qtpass.cpp` (`showInTextBrowser()`, `processErrorExit()`) each wrapped `protocolRegex()` matches in `<a href="...">` with no further check. The stricter `isLaunchableWebUrl()` predicate only gated the explicit open-in-browser button, so URL *detection* and the *click* path used different rules.

## Fix

- New `Util::linkifyUrls(const QString &text, bool *linked = nullptr)`: HTML-escapes the text and turns only those `protocolRegex()` matches that also pass `isLaunchableWebUrl()` (http/https, valid host, no userinfo, no control chars) into anchors. Everything else stays escaped plain text.
- `PasswordDisplayPanel::addField()` uses the helper (drops its private regex loop and the file-static regex copy); `linked` decides between `setPlainText()` and `setHtml()` as before.
- `QtPass::showInTextBrowser()` and `QtPass::processErrorExit()` use the helper instead of `toHtmlEscaped()` + regex replace.
- `protocolRegex()` itself is unchanged; the gate is the shared predicate, so the open-in-browser button and the inline anchors now agree.

## Tests

Added:
- `tst_util::linkifyUrlsLinksLaunchableWebUrls` — plain text escaped and not linked; `https://example.org/?a=1&b=2` produces exactly one correctly escaped anchor; two web URLs both link.
- `tst_util::linkifyUrlsLeavesNonWebUrlsAsText` — `ssh://user:pass@host`, `ftp://host`, `sftp://`, `webdav://`, `https://user:pass@…`, `http://user@…` come back as `toHtmlEscaped()` text with no anchor; mixed https + ssh links only the https URL.
- `tst_passworddisplaypanel::fieldValueWithNonWebUrlGetsNoAnchor` (data-driven: ssh with creds, ftp, sftp in prose, https with creds, webdav) — rendered text matches the value, no `<a href=`, no URL button.
- `tst_passworddisplaypanel::fieldValueLinksOnlyTheLaunchableUrl` — mixed value: https anchor present, ssh anchor absent, exactly one anchor.

Verified the panel tests fail on the previous `passworddisplaypanel.cpp` (6 failures) and pass with the fix. Ran `QT_QPA_PLATFORM=offscreen make -j2 check`: all suites 0 failed.

## Notes

- Behaviour change: non-web URLs (`ssh://`, `ftp://`, `webdav://`, …) and URLs with embedded credentials are no longer clickable anywhere; they still display as text and can be copied.
- Qt 5.15 backport: no Qt 6-only API used. `qsizetype` exists since Qt 5.10 and `QString::mid()` takes it by implicit narrowing.

Part of #1682; backport tracked in #1709.


annejan:
## Self-review pass

Verified the follow-up finding against the branch and fixed it in caf1604de.

### Fixed

**`Util::protocolRegex()` matched across newlines and tabs, so `linkifyUrls()` mishandled URLs at line ends** (`src/util.cpp`)

The URL character class `[^" <>\)\]\[]` excluded only a literal space. On the two multi-line callers of `Util::linkifyUrls()` (`QtPass::showInTextBrowser`, `QtPass::processErrorExit`) this meant:

- `url: https://example.org\nusername: foo` matched `https://example.org\nusername:`; `isLaunchableWebUrl()` rejected the embedded LF and the URL was never linked.
- `https://example.org\n` matched with the trailing newline; the predicate trimmed it and passed, but the emitted `href` kept the `\n`, which the callers' `\n` -> `<br />` replacement turned into `href="https://example.org<br />"`, so the URL handed to the OS handler was not the one that had been validated.

Fix: the class now excludes all whitespace (`[^"\s<>\)\]\[]`), matching the `\S` convention in `.opencode/skills/qtpass-fixing/SKILL.md`. With whitespace excluded from the capture, the validated string and the emitted `href` are always identical. Reproduced both cases with a probe and confirmed the new tests fail against the previous regex.

Tests added in `tests/auto/util/tst_util.cpp`:
- `utilRegexProtocol`: the match stops at `\n`, `\r\n` and `\t`.
- `linkifyUrlsLinksLaunchableWebUrls`: exact HTML for the multi-line pass-file case, a URL with a trailing newline, and a mixed CR/LF/tab sequence.

### Skipped

Nothing skipped. CodeRabbit did not review this PR (auto reviews are off for non-default base branches), so there were no bot findings to evaluate.

### Validation

`qmake6 qtpass.pro && make -j2 sub-src sub-main`, then `QT_QPA_PLATFORM=offscreen make -j2 check`: every suite 0 failed (tst_util 153 passed / 4 skipped, tst_passworddisplaypanel 40 passed). `clang-format` applied to both touched files.

PR #1721, opened 2026-09-14
Harden the single-instance IPC (#1682)

annejan:
## Summary

Hardens `SingleApplication`, the QLocalServer/QSharedMemory single-instance IPC behind `SINGLE_APP=1`, so a crashed instance can no longer disable it permanently, only the owning user can connect, accepted sockets are freed, and `main()` no longer exits silently when forwarding fails.

## Root cause

`src/singleapplication.cpp` called `localServer->listen(_uniqueKey)` and discarded the result. On Linux/macOS a socket file left in `$TMPDIR` by a crashed instance makes `bind()` fail with `EADDRINUSE`, so the new instance ran with no server. Every later launch attached the (also stale) shared-memory segment, took the `isRunning()` path, failed `sendMessage()` and hit `return 0;` in `main/main.cpp`, which ignored the `false`. Net effect: launcher clicks did nothing until the user deleted the socket by hand.

Secondary issues in the same code: the socket was created with the default umask (other local users could connect), `receiveMessage()` blocked the GUI thread in `waitForReadyRead(1000)` for every connection and never deleted the accepted `QLocalSocket`.

## Fix

- When the shared-memory segment attaches, probe the socket (`peerIsListening()`). If nobody answers the segment is stale (System V shm survives a crash on Unix): `detach()` and take over as the server instead of forwarding into the void forever.
- `QLocalServer::removeServer(_uniqueKey)` before `listen()`, `setSocketOptions(QLocalServer::UserAccessOption)`, and `qWarning()` if `listen()` still fails.
- `receiveMessage()` drains all pending connections and reads each asynchronously: bytes are buffered on `readyRead`, the message is emitted on `disconnected` (only if anything arrived, so a bare probe does not raise the window), and the socket is `deleteLater()`ed. No more blocking in the GUI thread.
- `main.cpp`: `if (app.isRunning() && app.sendMessage(text)) return 0;` falls through to a normal start when forwarding fails.

`QSharedMemory` is kept: its exclusive `create()` is what arbitrates two simultaneous launches, and with `UserAccessOption` Qt binds to a temp path and `rename()`s over the socket path, so a probe-only design would let a second instance silently replace a live peer's socket. The stale-segment problem it introduced is now handled by the probe/takeover.

All APIs used (`removeServer`, `setSocketOptions`, `QSharedPointer::create`, `QLocalServer::UserAccessOption`) exist in Qt 5.15, so the backport needs no alternatives.

## Tests

New suite `tests/auto/singleapplication/` (added to `auto.pro` under `!nosingleapp`). Because `SingleApplication` *is* the `QApplication`, the suite has its own `main()` that plants a stale file at the resolved socket path (Unix) and then constructs the application; tests talk to it with plain `QLocalSocket` clients like a second launch would:

- `listensDespiteStaleSocket` — a client can connect (fails before: "Connection refused")
- `socketIsUserAccessOnly` — socket has no group/other bits (fails before: `srwxr-xr-x`)
- `messageArrives` — text arrives via `messageAvailable`, accepted socket count drops to 0
- `emptyPayloadArrivesAsEmptyMessage` — the single-NUL payload of an argument-less launch surfaces as `""`
- `silentPeerIsIgnoredAndReleased` — connect + hang up: no signal, no leaked socket, no GUI stall
- `firstInstanceIsNotRunning` — `isRunning()` false, `sendMessage()` false

Ran: `qmake6 qtpass.pro && make -j2 sub-src sub-main`, then `QT_QPA_PLATFORM=offscreen make -j2 check` (Qt 6.11.2): all suites 0 failed. Against the unfixed `singleapplication.cpp` the new suite reports 5 failed / 3 passed.

Manual end-to-end with the built binary under an isolated `HOME`: second launch forwards and exits 0 with the first alive; after `kill -9` of the first, a third launch takes over the stale segment and socket and a fourth launch forwards to it.

## Notes

- Behaviour change: an instance that finds a stale segment now becomes the server; previously it exited without doing anything.
- Behaviour change: a connection that sends no bytes no longer emits `messageAvailable("")` (and so no longer raises the window). Real launches always send at least one byte.
- Not tested automatically: the stale-shared-memory takeover, because a System V segment with zero attachments cannot be produced in-process (Qt removes it on detach); it is covered by the manual run above.
- `.gitignore` gains the per-suite line for the new test binary, matching the existing convention.

Part of #1682; backport tracked in #1709.


annejan:
## Self-review pass

Two findings verified against the branch; both were valid and are fixed in db320878a.

### Fixed

**1. Failed `sendMessage()` left an orphan that blocked every later launch** (`main/main.cpp` fall-through / `src/singleapplication.cpp`)

Reproduced with a harness linked against this branch's `libqtpass.a`: A primary, B constructed while A is alive, A killed before B forwards. B fell through to a GUI start with `isRunning=1 hasServer=0` while still attached to the segment; launches C and D then each got `isRunning=0 hasServer=0` (attach ok, probe fails, `create()` fails because B holds the segment) and opened yet another window with no IPC, until B exited.

Fix: the constructor's `create(1)` / `removeServer` / `setSocketOptions` / `connect` / `listen` sequence is now a private `becomePrimary()`. `sendMessage()` calls it after a failed forward, after clearing `_isRunning` and detaching (the forwarding body moved to a private `forwardMessage()`). Same harness after the fix: B ends with `isRunning=0 hasServer=1` and receives the messages C and D forward. If the segment is still held (peer alive but unreachable), `create()` fails and the instance runs without IPC exactly as before, just no longer pinning the segment.

Test: `tst_singleapplication`'s `main()` now brings up a `FakePeer` (raw Unix socket + `QSharedMemory`, created before the application object since `QLocalServer` needs one) so the suite starts as a second instance. `takesOverWhenPeerVanishesBeforeForward()` makes the peer vanish (listen fd closed, socket file left behind as after a crash, segment released) and asserts `sendMessage()` returns false, `isRunning()` flips to false, a server exists and a client can connect. The other tests run against the taken-over instance; the stale-socket case is still covered because the fake peer's socket file is what `removeServer()` has to clear. On Windows there is no fake peer and the test skips. Fails 7/9 against the pre-fix source.

**2. `silentPeerIsIgnoredAndReleased` was vacuous** (`tests/auto/singleapplication/tst_singleapplication.cpp`)

Confirmed: `QTRY_COMPARE(acceptedSockets(), 0)` passed on its first evaluation before the server had accepted anything, so the released / no-stall / no-emit checks never saw the socket. The test now takes a `QSignalSpy` on the server's `newConnection`, waits for it (`accepted.wait(2000)`, `count == 1`), then runs the `QTRY_COMPARE`, the stall timer and the `messageAvailable` count. Verified by temporarily restoring the old blocking `receiveMessage()`: the test now fails (5 leaked sockets), whereas the previous version passed.

### Skipped

None.

### Validation

`qmake6 qtpass.pro`, `make -j2 sub-src sub-main`, `QT_QPA_PLATFORM=offscreen make -j2 check`: 27 suites, 951 passed, 0 failed (Qt 6.11.2). `clang-format` applied to the three touched files.

PR #1725, opened 2026-09-14
Backport: Pass --no-encrypt-to on every gpg encrypt call (#1709)

annejan:
## Summary

Backport of #1720 to `main` for the 1.8.1 patch release (#1709): every gpg encrypt call made by `ImitatePass` now carries `--no-encrypt-to` and `--compress-algo=none`, so an `encrypt-to <key>` line in the user's `gpg.conf` can no longer add a recipient that the (optionally signed) `.gpg-id` never listed.

## Original fix

`ImitatePass::Insert()` (`src/imitatepass.cpp`) and the per-file re-encryption in `reencryptSingleFile()` ran `gpg -eq` with only the `-r` recipients from `.gpg-id`. gpg still honours `encrypt-to` from `gpg.conf` on such calls, so every file written by QtPass would silently be readable by that extra key. `--no-encrypt-to` was only present on the decrypt calls, where it does nothing.

## Fix

Add `--compress-algo=none --no-encrypt-to` to both encrypt argv, exactly as pass(1) does through its `GPG_OPTS`. The flags on the decrypt calls are left in place (pass applies the same `GPG_OPTS` to decryption too; they are harmless there). The `src/` hunk cherry-picked cleanly.

## Tests

- `tests/auto/imitatepass/tst_imitatepass.cpp` (new suite on `main`): drives `Insert()` and `reencryptPath()` through a recording fake gpg and checks the encrypt argv (both flags present, the `.gpg-id` recipient still the only `-r`, insert/re-encrypt still succeed). Both tests fail on unfixed `main` (verified by rebuilding with the original `src/imitatepass.cpp`) and pass with the fix.
- `tests/auto/integration/tst_integration.cpp`: `imitatePass_insertIgnoresGpgConfEncryptTo` and `imitatePass_reencryptIgnoresGpgConfEncryptTo` add `encrypt-to <second key>` to the throwaway `gpg.conf` and verify with `gpg --list-packets` that Insert and re-encryption produce files targeting only the `.gpg-id` key while still decrypting to the inserted content. Applied unchanged.
- Ran locally on Qt 6.11.2: `QT_QPA_PLATFORM=offscreen make -j2 check` — all 25 suites pass, 0 failed.

## Notes

- Hand-port: the `tests/auto/imitatepass` suite does not exist on `main` (it was introduced by the 2.0-only re-encryption worker, #1697). This PR creates it with only the two encrypt-argv tests; the worker-thread tests were left out, and the re-encrypt test observes the synchronous `reencryptPath()` through plain `QSignalSpy` rather than queued connections. The suite is registered in `tests/auto/auto.pro` and its binary added to `.gitignore`.
- No new Qt 6-only API: `Qt::SkipEmptyParts` (5.14) is the newest thing used; no `QT_VERSION_CHECK` guards were needed.
- No behaviour change beyond the two extra gpg flags on encryption.

Backport of #1720 for #1709.

## Summary by CodeRabbit

- **Bug Fixes**
  - Encryption and re-encryption now use only the recipients specified for the password store, preventing additional GPG-configured recipients from being included.
  - Compression is consistently disabled during encryption, providing predictable behavior across workflows.

- **Tests**
  - Added automated coverage for recipient selection and encryption options during insertion and re-encryption.
  - Added integration tests covering interactions with GPG configuration, including configured encryption recipients.


annejan:
Self-review pass: the `QSaveFile` finding is valid per the project guideline (atomic writes) — `GpgConfLine` now commits the modified and the restored `gpg.conf` through `QSaveFile` in both constructor and destructor; `m_ok` is only true after `commit()`. Integration suite 29/29. Same change mirrored to `2.0-nazomer` (the helper came from #1715).

PR #1726, opened 2026-09-14
Backport: Only link launchable http(s) URLs in inline anchors (#1709)

annejan:
## Summary

Backport of #1719 (2.0-nazomer) to main for the 1.8.1 patch release: inline `<a href>` anchors in the password display panel and in `QtPass::showInTextBrowser()` / `processErrorExit()` are now only produced for URLs that pass `Util::isLaunchableWebUrl()`.

## Original fix

`Util::protocolRegex()` matches `ssh://`, `sftp://`, `ftp://` and `webdav://` as well as URLs carrying `user:pass@` credentials. All three anchor-producing sites wrapped every match in `<a href>`, and both `QTextBrowser`s call `setOpenExternalLinks(true)`, so a click handed a non-web scheme or embedded credentials straight to the OS URL handler, bypassing the gate the open-in-browser button already used.

## Fix

- `Util::linkifyUrls(text, bool *linked = nullptr)` (new): HTML-escape the text and turn only the `protocolRegex()` matches that also pass `isLaunchableWebUrl()` into anchors; everything else stays escaped plain text.
- `PasswordDisplayPanel::addField()`, `QtPass::showInTextBrowser()` and `QtPass::processErrorExit()` use it, so URL detection and gating share one predicate.
- `Util::protocolRegex()` now stops at any whitespace (`\s`) instead of only a literal space, so a URL followed by a line break in multi-line text (pass file body, gpg stderr) is captured without the `\n`, which the callers' `"\n" -> "<br />"` replacement would otherwise have folded into the `href`.

Two cherry-picked commits, `(cherry picked from commit …)` lines kept. Hand-port note in the first commit: main's `PasswordDisplayPanel` still renders through `QTextBrowser::setText()` (the `setPlainText()`/`setHtml()` split from #1683 is not on main), so the panel simply passes the value through `Util::linkifyUrls()` and keeps `setText()`. Nothing else in the fix needed adapting.

## Tests

- `tests/auto/util/tst_util.cpp`: `linkifyUrlsLinksLaunchableWebUrls`, `linkifyUrlsLeavesNonWebUrlsAsText`; `utilRegexProtocol` asserts the match stops at `\n`, `\r\n` and `\t`.
- `tests/auto/passworddisplaypanel/tst_passworddisplaypanel.cpp`: `fieldValueWithNonWebUrlGetsNoAnchor` (ssh with credentials, ftp, sftp in prose, https with credentials, webdav: no anchor, no URL button) and `fieldValueLinksOnlyTheLaunchableUrl` (mixed value links only the https URL). The `browserAt()` / `urlButtonAt()` helpers those tests use were carried over with them; the #1683/#1693 tests in the same 2.0 hunk were not, as they belong to other fixes.
- Ran locally on Qt 6: `tst_util` 140 passed / 0 failed / 4 skipped, `tst_passworddisplaypanel` 27 passed / 0 failed, and full `make check` (see below).

## Notes

- Behaviour change: `ssh://`, `sftp://`, `ftp://`, `webdav://` URLs and any URL with `user:pass@` are no longer clickable in the password pane or the text browser; they are still shown as text. `http(s)` URLs without credentials link as before.
- No Qt 6-only API in the diff (`qsizetype`, `QRegularExpressionMatchIterator`, `\s` in the regex are all available on 5.15); no `QT_VERSION_CHECK` guards were needed.
- No changes to localization/, docs/ or CHANGELOG.md.

Backport of #1719 for #1709.


annejan:
Before/after, rendered from the real `PasswordDisplayPanel` (offscreen) with a store entry containing web and non-web URLs:

main vs #1726: which values become clickable anchors

`<a href>` targets actually present in the rendered HTML:

| | main (1.8.0) | #1726 |
|---|---|---|
| `https://example.org/login` | anchor | anchor |
| `https://intranet.example.org/wiki` (inside a note) | anchor | anchor |
| `ssh://git:hunter2@gitea.example.org/vault.git` | **anchor** → OS `ssh` handler, credentials included | plain text |
| `sftp://backup.example.org/store` | **anchor** | plain text |
| `ftp://files.example.org/pub` | **anchor** | plain text |
| `ssh://ops:secret@bastion.example.org` (inside a note) | **anchor** | plain text |

Both text browsers have `setOpenExternalLinks(true)`, so on main a click on any of the bold rows went straight to `QDesktopServices::openUrl()`.

PR #1728, opened 2026-09-14
Backport: Harden the single-instance IPC (#1709)

annejan:
## Summary

Backport of the single-instance IPC hardening (#1721) from 2.0-nazomer to main for the 1.8.1 patch release (#1709). Both commits cherry-pick cleanly onto main; `src/singleapplication.*` is identical between the two branches, so the code lands verbatim.

## Original fix

A socket file left behind by a crashed QtPass made `QLocalServer::listen()` fail with "address in use". The return value was ignored, so the new instance ran without a server, and every later launch silently exited after a failed `sendMessage()`. Launcher clicks did nothing until the user deleted the socket file by hand. Related: the stale System V shared-memory segment outlived the crashed process, so the new instance believed a peer was running and forwarded into the void; and when the peer vanished between the probe and the forward, `main()` fell through while still attached, so every later launch failed `create()` and opened another IPC-less window.

## Fix

- Probe the socket when the shared-memory segment attaches; when nobody answers the segment is stale, detach and take over.
- `QLocalServer::removeServer()` before `listen()`, set `UserAccessOption` so only the owning user can connect, and `qWarning()` when `listen()` still fails.
- Accepted sockets are read asynchronously; an empty connection is not reported as a message, and every accepted `QLocalSocket` is deleted once the peer disconnects (no more 1 s GUI stall on a silent peer).
- `sendMessage()` that fails clears `_isRunning`, detaches, and calls the factored-out `becomePrimary()`, so the window `main()` opens next is reachable. `main()` only exits early when `isRunning() && sendMessage(text)` both succeed.

## Tests

- New `tests/auto/singleapplication/tst_singleapplication` (comes with the cherry-pick, added to `auto.pro` under `!nosingleapp`): a raw-POSIX fake peer holds the segment and socket, vanishes before the forward, and the suite verifies takeover over the stale socket file, 0700 socket permissions, message and empty-payload delivery, silent-peer handling without stall or leak, and `isRunning() == false` for the primary.
- Ran locally on Qt 6.11.2: `tst_singleapplication` 9 passed / 0 failed; full `QT_QPA_PLATFORM=offscreen make -j2 check` all suites 0 failed.
- Extra commit for Windows CI: the client side of `messageArrives`/`emptyPayload` blocked in `waitForBytesWritten()`, which never runs the event loop; the server under test is in the same process and only accepts and reads from that loop. On Windows `QLocalServer` uses zero-sized pipe buffers, so the write could not complete and the test timed out (and the un-accepted connections cascaded into `silentPeerIsIgnoredAndReleased`). The test now spins the loop until `bytesToWrite()` is 0. Unix behaviour is unchanged. #1721 never ran the build workflow, so 2.0-nazomer needs this fix too.
- Qt 5.15: the diff uses no Qt 6-only API (`QLocalServer::UserAccessOption`, `QSharedPointer::create`, `QScopedPointer::reset`, `QFileInfo::exists(path)` all exist in 5.15). CI matrix will confirm.

## Notes

- Behaviour change: a launch whose forward fails now opens a window (and becomes the listening instance) instead of exiting silently.
- Only conflict during cherry-pick was a `.gitignore` context line (`tests/auto/imitatepass` exists only on 2.0); resolved by adding just the `tst_singleapplication` entry. No code was adapted.

Backport of #1721 for #1709.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved single-instance handling when an existing instance becomes unavailable.
  - Failed message delivery no longer causes the launcher to exit silently; a new instance can start when appropriate.
  - Stale single-instance state is detected and recovered automatically.
  - Message reception remains responsive when another instance connects without sending data.
  - Improved handling of empty messages and secure local communication.


annejan:
Self-review pass on the CodeRabbit findings, each verified against the code:

- **Unbounded per-connection buffer** — valid (same-user only, socket is `0700`, still worth a cap). `readyRead` now aborts the peer once 64 KiB is exceeded; new test `oversizedPayloadIsDroppedAndPeerDisconnected`.
- **Takeover race** — valid: two launchers recovering from one crash both detach, the loser of `create(1)` ran as a second full instance. `becomePrimary()` now attaches to the winner's segment and probes it (5 × 100 ms) and becomes its secondary, so `main()` forwards.
- **`listen()` fails while holding the segment** — valid: later launches attached, failed the probe, then failed `create()` forever. The failure path now resets the server and detaches.
- **Test order dependence** — valid: `tst_singleapplication messageArrives` alone failed (`spy.wait(2000)` returned FALSE, talking to the fake peer). `ensurePrimary()` at the top of every test that needs the listening instance; single-function runs pass now.
- **`qExec(QCoreApplication::arguments())`** — applied.

Full suite 10/10, single-function runs pass, desktop crash/relaunch scenario re-run with this build (relaunch after `kill -9` opens a window, next launch forwards and exits). Same fix goes to `2.0-nazomer` as a follow-up to #1721.

PR #1735, opened 2026-09-14
Fix copying an entry onto a folder in the native backend (#1682)

annejan:
## Summary

Ctrl-drag copying an entry onto a folder in the tree always failed with "Copy failed" when QtPass uses its native (ImitatePass) backend. `ImitatePass::Copy` now resolves a folder destination the same way `ImitatePass::Move` already does, so the copy lands as `<folder>/<entry>.gpg`.

## Root cause

`StoreModel::executeDropAction` hands the *folder* path to the backend for a file-onto-folder drop (for both move and copy). That is the contract the pass CLI backend relies on: `RealPass::passMoveOrCopy` runs `pass cp/mv <name> <folder>`, and pass resolves the target itself. `ImitatePass::Move` honours the same contract through `resolveMoveDestination()` (dir/file/`.gpg` handling, force/clash rules). `ImitatePass::Copy` did not: it called `QFile::copy(src, dest)` with the folder as `dest`, which cannot write a directory, so `QFile::copy` returned false and the "Copy failed" critical fired on every copy onto a folder.

## Fix

`src/imitatepass.cpp`, `ImitatePass::Copy` only:

- Resolve the target with the existing `resolveMoveDestination(src, dest, force)` (reused, not duplicated) and use the resolved path for `QFile::remove` (force), `QFile::copy`, `git add`, the commit message and the re-encryption.
- When `resolveMoveDestination` refuses (existing target without force, missing source), emit the existing "Copy failed" critical rather than returning silently.

`StoreModel` keeps handing the folder path to the backend and `RealPass::Copy` is untouched, so the pass CLI backend is unaffected (this is where the closed #1734 went wrong).

## Tests

Added to `tests/auto/integration/tst_integration.cpp`, next to `imitatePass_insertCopyAndShow`:

- `imitatePass_copyOntoFolderAndShow` — `Copy()` an existing entry onto a *folder path*; asserts no critical, source untouched, `<folder>/<entry>.gpg` exists (and no doubled `.gpg`), and `Show("folder/original")` decrypts the copied content.
- `imitatePass_copyOntoExistingEntryClash` — `Copy()` onto an existing entry: `force=false` emits exactly one "Copy failed" critical and leaves both files byte-identical; `force=true` overwrites the target, which then decrypts to the copied content.

On `origin/2.0-nazomer` (test file applied, source unchanged) the new folder test fails:

```
FAIL!  : tst_integration::imitatePass_copyOntoFolderAndShow() 'criticalSpy.isEmpty()' returned FALSE. (copy onto a folder must not fail)
   Loc: [tst_integration.cpp(678)]
```

(The clash test already passes on the base; it pins the user-visible critical for the new early-return path.)

With the fix, run with `QT_QPA_PLATFORM=offscreen`:

- `tests/auto/integration/tst_integration`: 37 passed, 0 failed
- `tests/auto/imitatepass/tst_imitatepass`: 7 passed, 0 failed
- `tests/auto/model/tst_model`: 38 passed, 0 failed (drop test still asserts the folder-path contract)

## Notes

- Behaviour change for a non-existent destination without a `.gpg` suffix: `Copy(src, "<store>/x")` now creates `<store>/x.gpg` (as `Move` already does) instead of a bare `x`. The only caller (`StoreModel`) always passes `.gpg` names or folders.
- Copying a whole folder still fails as before (`QFile::copy` does not copy directories); that is a separate item.
- `clang-format` applied; Qt 5.15-compatible API only.

Part of #1682. Supersedes the closed #1734 for this item.


annejan:
Self-review pass: the reviewer's low-severity note that `Copy(src, <src's own folder>, force=true)` removes the source was checked and is a real data-loss path (`resolveMoveDestination` maps that folder to `src`, the forced `QFile::remove` deletes it, the copy then fails). Guarded with the existing "Copy failed" critical; the clash test now covers it and fails without the guard. Not reachable from drag-and-drop (`force=false` there), but `Copy()` is public. The other two notes (folder-destination clash coverage, folder *copy* still unsupported in the native backend) are pre-existing scope; the second stays on #1682.

PR #1748, opened 2026-09-15
Backport: fix copying an entry onto a folder in the native backend (#1735)

annejan:
Backport of #1735 to the 1.8 line, plus the follow-up from #1746 (`destFileInfo.refresh()` after the copy, with an integration test).

## What it fixes

Ctrl-drag copying an entry onto a folder in the tree always failed with "Copy failed" when QtPass uses its native (gpg2/git) backend. `StoreModel::executeDropAction` hands the *folder* path to the backend, as `pass cp` expects; `ImitatePass::Move` resolves that through `resolveMoveDestination()`, `ImitatePass::Copy` passed the folder straight to `QFile::copy`, which cannot write a directory. Copy now resolves the destination the same way Move does (into the folder, `.gpg` appended, no clobbering without force, refuses a copy onto the entry's own folder so `force` cannot delete the only copy).

## Commits

- `ccbea23df` cherry-pick of #1735 (clean)
- `a9491d130` cherry-pick of 9d1e23ddf from #1746: refresh the cached `QFileInfo` before choosing the re-encryption path, new test `imitatePass_copyIntoFolderReencryptsToFolderKey`
- `559eae9bf` 1.8-only test adjustment: the clash test compared the source's bytes after a forced copy; on 1.8 `Copy` re-encrypts the folder synchronously (2.0 does it on a worker thread), so the source ciphertext is legitimately rewritten. The test now checks the source survives and the target changed; the decrypt check that follows covers the content.

## Validation

`tests/auto/integration`: 35 passed, 0 failed (Qt 6.11.2, temporary GNUPGHOME). Full `make check` running, result posted below.

For 1.8.1; CHANGELOG entry comes with the release PR.


annejan:
Full `make check` on the branch (Qt 6.11.2, offscreen): 27 suites, 966 passed, 0 failed.

PR #1752, opened 2026-09-16
Remove WebDAV mounting (#1682)

annejan:
Section 1 item of #1682: **Delete WebDAV mounting** — `src/settingsserializer.cpp:167`, `src/qtpass.cpp:215-283`.

## Why

`QtPass::mountWebDav()` mounted the password store from a WebDAV server: on Windows through `WNetUseConnectionA`, elsewhere by starting `fusedav`.

- No user interface exists, and never has: `useWebDav`, `webDavUrl`, `webDavUser` and `webDavPassword` can only be set by editing the configuration file by hand.
- The password is stored there in plain text (mode 0644 ini on Linux).
- The `fusedav` invocation passes `"\"" + user + "\""` and `"\"" + passStore + "\""` as single argv elements, so the quotes reach `fusedav` literally — broken since the shell-less `QProcess` argument list landed in 2020.
- `useWebDav` also silently disabled Git in three places in `ImitatePass`, which is how a store on a WebDAV mount avoided Git operations.

## What changes

- `src/qtpass.{h,cpp}`: `mountWebDav()`, the `fusedav` `QProcess`, the Windows disconnect in the destructor (now `= default`), the mount call in `init()`, and the `<winnetwk.h>` / `QInputDialog` include split are gone.
- `src/appsettings.h`, `src/settingsconstants.{h,cpp}`, `src/qtpasssettings.{h,cpp}` (`isUseWebDav`), `src/settingsserializer.cpp`: the four settings removed.
- `src/settingsserializer.cpp`: saving now `remove()`s the four obsolete keys, so a configuration written by 1.8.x loses its plaintext `webDavPassword` the first time 2.0 saves.
- `src/imitatepass.cpp`: `!m_settings.useWebDav && gitReady()` → `gitReady()` (3 sites).
- Docs: the feature claim drops out of `README.md`, `publiccode.yml` and `qtpass.appdata.xml`; the README's "one OTP/U2F enabled WebDAV account per password" smartcard workaround goes too; `FAQ.md` gains **What happened to WebDAV support?** pointing at Git or a folder the operating system already syncs/mounts (Nextcloud, Syncthing, an SMB or WebDAV share) — those clients handle credentials, reconnects and conflicts, which the removed code did not.
- `CHANGELOG.md`: entry under 2.0 → Removed.

`localization/*.ts` is deliberately untouched; the removed `tr()` strings become obsolete entries at the next `lupdate` run (same convention as #1737).

## Tests

- New `tst_settings::serializerDropsObsoleteWebDavKeys()`: writes a 1.8-style ini with all four keys, loads and saves it, then asserts the keys are gone, that the literal password string is absent from the file, and that an unrelated key survives.
- The four WebDAV rows are removed from the round-trip and key tables.
- `QT_QPA_PLATFORM=offscreen make check`: **27 suites, 983 passed, 0 failed.**
- `appstreamcli validate` OK, `publiccode.yml` parses, super-linter (markdown, textlint, xml, yaml) clean.

## Summary by CodeRabbit

- **Removed Features**
  - Removed experimental WebDAV mounting and synchronization support.
  - Existing WebDAV settings, including the stored password, are deleted when settings are saved.
  - Git-based synchronization now works independently of WebDAV settings.

- **Documentation**
  - Updated the FAQ and product documentation to explain the removal and recommend Git or system-synced folders as alternatives.
  - Removed WebDAV references from feature listings and security guidance.


annejan:
Verified against the code on this branch: the guard at `src/imitatepass.cpp:229` already requires Git to be enabled, so no change is needed there.

```cpp
auto ImitatePass::gitReady() -> bool {
  if (m_settings.useGit && m_settings.gitExecutable.isEmpty()) {
    emit statusMsg(tr("Git executable not configured, skipping git"), 3000);
    return false;
  }
  return m_settings.useGit;
}
```

`gitReady()` returns `m_settings.useGit` — the alternative the comment suggests ("enforce this requirement within `gitReady()` itself") is what it already does. `QtPassSettings::isUseGit()` would read the same setting from the singleton instead of the `AppSettings` copy `Pass::init()` was given; every other backend call in this file uses the copy, so adding it would mix the two sources without changing behaviour.

What this PR removed from the condition is only the `!m_settings.useWebDav` term, which disabled Git when the store was a WebDAV mount; with the mount gone that term can never be true.

Added a regression test anyway, 7e1d26953: `tst_imitatepass::insertRunsNoGitWhenGitIsDisabled()` runs `Insert()` with `useGit = false` and a *configured* recording fake git, then asserts the git log file was never created. `imitatepass`: 13 passed, 0 failed.

annejan:
Both comments handled in 9701e428c.

**`tests/auto/imitatepass/tst_imitatepass.cpp:690` — the 300 ms wait.** Valid point, fixed without a sleep. `Insert()` queues its git calls in the same call as the gpg one, and `Executor` runs a single FIFO queue (`m_execQueue`, one `running` task at a time), so the test now queues a *second* `Insert()` and waits for its `finishedInsert`; everything the first one queued has necessarily run by then. Ran it 10× in a row, stable, and the whole suite is at 27 suites / 984 passed / 0 failed.

**`src/imitatepass.cpp:229` — the `isUseGit()` guard.** Still not a code change: `gitReady()` returns `m_settings.useGit`, so the condition the guideline asks for is exactly what it enforces. What was actually wrong is the guideline itself — `AGENTS.md` predates `gitReady()`/`gitConfigured()` (#1691, #1716) and still tells contributors to call `QtPassSettings::isUseGit()` from a backend, which reads the settings singleton while the rest of `ImitatePass` reads the `AppSettings` copy handed to `Pass::init()`. That line is now replaced by a description of the two helpers and why a backend must not reach for the singleton, so the rule matches the code that the code-review bot is checking against.

PR #1753, opened 2026-09-16
Make lupdate an explicit target instead of a configure-time side effect (#1682)

annejan:
Section 7 item of #1682: **Make `lupdate` an explicit target** — `qtpass.pro:62`.

## Problem

```qmake
system($$QMAKE_LUPDATE -locations absolute ./src ./main -ts $$files(localization/*.ts))
```

A `system()` call at the top level of a `.pro` file runs on **every** qmake invocation, including the one `make` triggers for a sub-project and the one CI runs before building. It rewrites all 64 `.ts` files (source line numbers included), so:

- a plain `qmake6` leaves `localization/` dirty, and contributors learn to `git checkout -- localization/` reflexively;
- `.github/workflows/freebsd.yml` had that workaround baked in;
- `Windows.md` documented the pollution as a fact of life;
- it has already required two whole-repository cleanup commits (`6a4f5e84a`, `11bc76099`, −22k lines) when the junk was committed by accident.

## Change

```qmake
lupdate.target = lupdate
lupdate.commands = $$QMAKE_LUPDATE -locations absolute $$PWD/src $$PWD/main -ts $$files($$PWD/localization/*.ts)
lupdate.depends = FORCE
QMAKE_EXTRA_TARGETS += lupdate
```

Same command, same arguments, same `-locations absolute` (Weblate sees no format change) — it just needs asking for now:

```sh
make lupdate      # nmake lupdate on Windows
```

Paths are `$$PWD`-anchored so the target works from a shadow build too.

Docs updated to match: `CONTRIBUTING.md` (adding a language), `Windows.md` (the note now says a plain qmake leaves the files alone), `.opencode/skills/qtpass-localization/SKILL.md` (three places), and the FreeBSD workflow loses its `git checkout -- localization/` line.

## Validation

- `qmake6 CONFIG+=release` on a clean tree → `git status` shows **no** modified `.ts` files (before: 64).
- `make lupdate` → "Updating 'localization/localization_*.ts'… 338 source texts" for all 64, exactly as the old configure step did.
- Full `QT_QPA_PLATFORM=offscreen make check`: **27 suites, 990 passed, 0 failed**, `localization/` still clean afterwards.


annejan:
zizmor 1.25.2 (new `ref-version-mismatch` audit in the super-linter image) flagged `.github/workflows/freebsd.yml:27`: the pin `vmactions/freebsd-vm@8b0f1a8f…` carried `# v1`, a moving major tag, while that SHA is the `v1.5.6` release. Comment corrected in the last commit — same SHA, no behaviour change. Local super-linter run with `VALIDATE_GITHUB_ACTIONS_ZIZMOR`: clean.

It is a pre-existing finding on `main`, not caused by this PR, but this PR touches that file so it may as well carry the fix.

PR #1755, opened 2026-09-16
Make qtpass.spec build this tree again (#1682)

annejan:
Section 7 item of #1682: **`qtpass.spec` cannot build this tree** — `qtpass.spec:17`, `:37`.

## What was broken

- `BuildRequires: qt5-qtbase-devel` / `qt5-linguist` and `%build: qmake-qt5`, against a `qtpass.pri` that `error()`s out on Qt 5 since 2.0. The spec could not produce a package at all.
- `qmake-qt5 PREFIX=%{buildroot}/%{_bindir}` put the binary in `/usr/bin/bin/qtpass` and baked the buildroot into the install prefix.
- `License: GPLv3` (not SPDX), `Requires: pass` (QtPass drives `gpg`/`git` directly by default), `Requires: qt6-qtbase` hand-written next to Qt 5 build deps.
- The icon and desktop file were installed by hand because `make install` only produced the binary — no longer true since the install rules landed.

## Now

- Qt 6 build deps (`qt6-qtbase-devel`, `qt6-qtsvg-devel`, `qt6-qttools-devel`), qmake through `%qmake_qt6` / `%qmake6` when the distribution defines one, `qmake-qt6` otherwise.
- `PREFIX=%{_prefix}` + `%make_install INSTALL_ROOT=%{buildroot}`; the desktop file, AppStream metainfo, both hicolor icons and the man page come from the build.
- `License: GPL-3.0-or-later`; `pass`, `gnupg2` and `git` are `Recommends`.
- A `%check` that runs `desktop-file-validate` and `appstream-util validate-relax`.
- `Release: 1` for 2.0.0 with a `%changelog` entry; the old entries kept (two had weekday/date mismatches that `rpmspec` warned about — fixed).

## Validation

`rpmbuild --nodeps -bb` on a `git archive` tarball of this branch (openSUSE, so the Fedora-named BuildRequires are skipped with `--nodeps`; the build itself is distribution-independent): builds, `%check` passes, payload is exactly

```
/usr/bin/qtpass
/usr/share/applications/qtpass.desktop
/usr/share/icons/hicolor/512x512/apps/qtpass-icon.png
/usr/share/icons/hicolor/scalable/apps/qtpass-icon.svg
/usr/share/man/man1/qtpass.1.gz
/usr/share/metainfo/qtpass.appdata.xml
+ LICENSE, README.md, CHANGELOG.md, FAQ.md
```

with auto-generated `libQt6*` dependencies and no unpackaged files. `rpmspec -P` is warning-free.

Note: the `BuildRequires` names are Fedora's, as the file header has always said; openSUSE spells them `qt6-base-devel` / `qt6-svg-devel` / `qt6-tools-devel`, noted in a comment.

Depends on nothing, but reads better after #1751 (which added the install rules this relies on).

## Summary by CodeRabbit

* **Chores**
  * Updated the RPM package to build QtPass 2.0.0 with Qt 6.
  * Updated runtime requirements for Qt SVG and GnuPG 2.
  * Made `pass` and `git` recommended rather than required packages.
  * Improved installation of application resources and package validation.
  * Added licenses, documentation, icons, metadata, and the man page to the package.
  * Removed legacy Qt 5 packaging steps and icon-cache scripts.


annejan:
Valid, applied in ec750e0d4: `Recommends: gnupg2` → `Requires: gnupg2`.

Checked against the code rather than assumed: both backends decrypt through gpg (`ImitatePass` runs it directly, `RealPass` shells out to `pass`, which runs it), and `Util::configIsValid()` (`src/util.cpp:244`) refuses a configuration whose `gpgExecutable` is missing — the first-run wizard then keeps re-asking. So a QtPass install without gnupg cannot open a store at all.

`pass` and `git` stay `Recommends`: the pass CLI is one of two backends (the default one talks to gpg directly), and git only syncs the store. Rebuilt with `rpmbuild --nodeps -bb`; the package now carries `Requires: gnupg2` alongside the auto-generated `libQt6*` dependencies.

PR #1757, opened 2026-09-16
Build resources reproducibly (rcc --format-version 1)

annejan:
Adopts Debian's `02-make-reproducible.patch` upstream (carried by the Debian package since 2019, `Forwarded: not-needed` — it should have been).

## Why

`rcc`'s default resource format writes each input file's modification time into the generated `qrc_*.cpp`, so two builds of the same source tree are not bit-identical. `--format-version 1` omits the timestamps.

Measured on `main`:

```
build 1: 096ab88509e7e896271937a9086f30db  src/qrc_resources.cpp
touch icons/*.svg artwork/*.png
build 2: 6eca05353f90767ffa7a36739d014d5a  src/qrc_resources.cpp   ← differs
```

With the flag, both builds give `097836d027cdf2fc7c44981d2cb3105f`.

## Change

One line in `qtpass.pri`, so it applies to `src` (the `resources.qrc` and the generated `qmake_qmake_qm_files.qrc` holding the translations) and to any future resource:

```qmake
QMAKE_RESOURCE_FLAGS += --format-version 1
```

Confirmed in the generated makefile:

```
/usr/libexec/qt6/rcc --format-version 1 -name resources ../resources.qrc -o qrc_resources.cpp
/usr/libexec/qt6/rcc --format-version 1 -name qmake_qmake_qm_files qmake_qmake_qm_files.qrc -o qrc_qmake_qmake_qm_files.cpp
```

`QT_QPA_PLATFORM=offscreen make check`: 27 suites, 984 passed, 0 failed — the resource system behaves identically, format 1 is what Qt used before the timestamped format and is still fully supported by `QResource`.

Part of reducing the Debian delta; see the packaging notes in the linked issue.

## Summary by CodeRabbit

- **Documentation**
  - Added guidance for Debian-based downstream releases, including package coordination, synchronization, signed assets, issue reporting, and patch maintenance.

- **Build Improvements**
  - Updated resource generation to produce reproducible output by excluding file modification timestamps, helping ensure consistent builds across environments.


annejan:
Both valid, fixed in fb0041fde.

**Local path (line 173).** `~/debian-qtpass-1.8.1-bug.txt` only exists on my machine — useless to anyone else reading the skill. The section now carries the report skeleton inline (pseudo-headers plus what each paragraph has to contain) and keeps the pointer to #1682 for the 1.8.1 specifics.

**Ubuntu/Mint routing (lines 163-165).** "Ubuntu syncs from Debian … the only lever" was too flat. Checked before rewriting:

- Ubuntu's auto-sync from Debian unstable runs only while the merge window is open (until DebianImportFreeze); after that it takes an explicit `requestsync` or a manual upload, and past FeatureFreeze a Feature Freeze Exception as well.
- Ubuntu does carry its own delta when it wants one: `qtpass 1.4.0-3ubuntu1` (plucky) added `Provides: pinentry`, and `1.4.0-3ubuntu2` is a no-change rebuild for 26.04. So an Ubuntu-side upload is possible — just more work, and repeated every cycle.
- Linux Mint's main editions are rebuilt from Ubuntu, but **LMDE** is built from Debian stable and follows Debian directly. That distinction was missing.

The conclusion is unchanged (Debian is where to coordinate — one upload eventually reaches all of them, an Ubuntu-only fix reaches neither Debian nor LMDE), but it is now stated with the reason instead of asserted.

Prettier and super-linter clean.

PR #1761, opened 2026-09-16
Raise the declared Qt floor to 6.8, the oldest version CI builds (#1682)

annejan:
Replaces the section 7 item **"Add the declared Qt 6.2 floor to CI (or raise the floor honestly)"** of #1682 with the second option.

## Why the honest option

Nothing older than **6.8** has ever been compiled:

| where | Qt |
|---|---|
| `ccpp.yml` matrix | 6.8 (LTS) on Linux + Windows, 6.11 on Linux/Windows/macOS |
| `release-installers.yml` | 6.8 (Windows), 6.11 (macOS) |
| Flatpak runtime | 6.11 |
| FreeBSD port | current `qt6-*` packages |

A floor that is declared but never built is not a supported configuration, it is an untested promise — and it has now broken twice:

- `tst_locale` avoids `QCOMPARE_NE` (Qt 6.4) with a comment explaining that the floor is 6.2;
- last week a `QBasicTimer::id()` / `QTimerEvent::id()` pair (Qt **6.8**) reached #1760 and nothing caught it, because every build in the matrix is ≥ 6.8. It was only found in review.

Adding a 6.2 job instead would mean building a version no user of ours runs (6.2 is out of support; distributions shipping Qt 6 are on 6.4+, Debian trixie on 6.8), for the sole purpose of keeping a promise nobody asked for.

## Changes

`qtpass.pri`'s guard (`lessThan(QT_MINOR_VERSION, 8)`), `README.md`, `CONTRIBUTING.md`, `AGENTS.md`, `CLAUDE.md`, `main/main.cpp`'s header comment, `publiccode.yml` (`versionMin: "6.8"`), and the 2.0 changelog entry. The `qtpass-fixing` skill's "Qt Version Floor" section now says the floor tracks the CI matrix and why. `tst_locale` keeps the `QVERIFY` workaround but loses the stale justification.

No code changes: nothing in `src/` used a 6.2–6.7-only API, and #1760 already replaced the 6.8-only timer API with `QTimer`, so this is purely the declaration catching up with reality.

## Validation

`QT_QPA_PLATFORM=offscreen make check`: **27 suites, 983 passed, 0 failed** on Qt 6.11.2. super-linter (markdown, textlint, yaml) clean.

## Summary by CodeRabbit

- **Documentation**
  - Updated build, contribution, and compatibility guidance to require Qt 6.8 or newer.
  - Clarified that Qt 6.8 is the oldest version covered by continuous integration.
  - Updated the changelog and project metadata to reflect the new minimum Qt version.
  - Clarified compatibility guidance for APIs introduced after Qt 6.8.
  - Qt 5 and Qt versions older than 6.8 are no longer supported.


annejan:
Valid, fixed in 96d64b3e7. The matrix runs 6.11 on `ubuntu-latest` and `macos-latest`; Windows is covered by the 6.8 entry only, so "6.11 on all three" was wrong. The section now names the two platforms and records why macOS has no 6.8 job (the runner's Xcode cannot compile `qyieldcpu.h` before 6.11.1, which is already documented in the workflow itself).

Coverage per platform, for the record: Linux 6.8 + 6.11, Windows 6.8, macOS 6.11, FreeBSD current qt6 in its own workflow.

PR #1762, opened 2026-09-16
Rebuild the window only when always-on-top changes; one Ui ownership idiom (#1682)

annejan:
Two items of #1682: **Window flags rebuilt on every Settings OK** (section 2, `src/mainwindow.cpp:474`) and **Unify Ui ownership; fix the `ImportKeyDialog` leak** (section 4, `src/importkeydialog.h:73`, `src/mainwindow.h:297`).

## Window flags

```cpp
if (QtPassSettings::isAlwaysOnTop()) {
  setWindowFlags(windowFlags() | Qt::WindowStaysOnTopHint);
} else {
  setWindowFlags(Qt::Window);      // ← every Settings OK, for everyone with the box unticked
}
show();
```

`setWindowFlags()` on a top-level widget goes through `setParent(nullptr, flags)`, i.e. it destroys and recreates the native window. The "off" branch ran that on **every** Settings OK, because `Qt::Window` never equals the current flags. That rebuild is the one `focusInput()` documents as having left `ui->lineEdit` dangling under Qt 6.11 (the `findChild` workaround stays; it is still correct defence).

Now: compare the one hint against the setting, return if unchanged; toggle only that hint (the old reset to `Qt::Window` also dropped any other flag); re-`show()` only if the window was visible — the first-run path calls this from the constructor before the window is ever shown.

`restoreWindow()` had a second copy of the logic that could **add** the hint at startup but never remove it; it calls the shared function now.

**Wayland**: `Qt::WindowStaysOnTopHint` is a no-op there (stacking belongs to the compositor). The checkbox is disabled with a tooltip pointing at the compositor's "keep above" window rules, same pattern as the tray checkbox when no tray exists.

## Ui ownership

Three idioms coexisted:

| class | before |
|---|---|
| `ConfigDialog`, `MainWindow` | `QScopedPointer<Ui::…>` |
| `ExportPublicKeyDialog`, `KeygenDialog`, `UsersDialog`, `PasswordDialog` | raw pointer + `delete ui` in the destructor |
| `ImportKeyDialog` | raw pointer, `~ImportKeyDialog() = default` → **one `Ui` struct leaked per import** |

All seven are `QScopedPointer<Ui::…>` now, with an out-of-line `= default` destructor so `Ui::X` is complete at the deletion point. `MainWindow::config()` also constructs its `ConfigDialog` on the stack like every other dialog caller instead of via a `QScopedPointer` — that was the "make the ConfigDialog a stack object like its siblings" part of the item.

## Tests

- `tst_mainwindow::windowFlagsAreOnlyRebuiltWhenAlwaysOnTopChanges` — shows the window, captures `winId()`, re-applies "off" twice → same `winId`, still visible; turns it on → hint set, window still shown; re-applies "on" → same `winId`; turns it off → hint cleared, `Qt::Window` kept. **Fails on the old code** (`restoreWindow()` never cleared the hint).
- `tst_importkeydialog::destroyingTheDialogDestroysItsWidgets` — `QPointer` to a Ui-created child goes null when the dialog goes out of scope.

`QT_QPA_PLATFORM=offscreen make check`: **27 suites, 989 passed, 0 failed.**

## Summary by CodeRabbit

- **New Features**
  - On Wayland, the “Always on top” option is disabled and directs users to configure this behavior through their compositor.

- **Bug Fixes**
  - Improved window-state handling when enabling or disabling “Always on top,” preserving other window settings and avoiding unnecessary window redisplays.
  - Window startup restoration now applies the saved always-on-top preference without forcing the window to appear.

- **Tests**
  - Added coverage for dialog cleanup and window-state restoration, including repeated settings changes.


annejan:
Leak claim verified with AddressSanitizer rather than asserted (`qmake6 "CONFIG+=debug sanitizer sanitize_address"`, `ASAN_OPTIONS=detect_leaks=1`, the `importkeydialog` suite):

**`main`:**
```
Direct leak of 88 byte(s) in 1 object(s) allocated from:
    #0 … in operator new(unsigned long)
    #1 … in ImportKeyDialog::ImportKeyDialog(QString const&, QWidget*) src/importkeydialog.cpp:28
…
SUMMARY: AddressSanitizer: 440 byte(s) leaked in 5 allocation(s).
```
One `Ui::ImportKeyDialog` (88 bytes) per dialog constructed — five dialogs in the suite, five leaks, exit code 1.

**This branch:** exit code 0, no leak report.

annejan:
Both checks addressed in 8cfbd44f6.

**Lint** — codespell: `onTop ==> on top` on a local variable in the new test. Renamed to `afterEnable`.

**codecov/project** (`68.89%, -0.05%`) — the interesting one. The three uncovered lines are the new Wayland branch in `ConfigDialog` (`platformName().startsWith("wayland")` → disable the checkbox), which the offscreen QPA can never execute. That is precisely the run-to-run noise `.codecov.yml` says it tolerates with a 0.2% threshold. It was failing anyway because **the file is invalid**: codecov's validator rejects it (`Error at ['coverage', 'patch']: unknown field`) — `project:`/`patch:` sit directly under `coverage:` instead of under `coverage: status:`, so since #1365 codecov has been silently applying its default 0% threshold. Fixed the nesting; the validator now returns `Valid!` with `threshold: 0.2`. Codecov reads the yml from the head commit, so this PR's re-run should pass, and so should every future PR that wobbles by a few lines.

annejan:
Correct, and I should have caught it myself: `inputTextEdit` is a child of the dialog, so Qt deletes it with the dialog whether or not the `Ui::ImportKeyDialog` allocation is freed — the test passed on the leaking code too, and its comment claimed more than it observed. Removed in 724e8e80b rather than kept with a softer comment; a test that cannot fail for the bug it is named after is worse than none.

The leak is pinned the way the comment suggests, by the sanitizer run posted above: `main` reports `5 × 88 byte(s)` from `importkeydialog.cpp:28`, this branch is clean. #1763 puts that recipe (`qmake6 "CONFIG+=debug sanitizer sanitize_address"`, `ASAN_OPTIONS=detect_leaks=1 make check`) into the testing skill with the clean full-suite baseline, so it is the standing regression check for ownership changes rather than a one-off.

PR #1765, opened 2026-09-16
skills: the BSD port test bed is where port patches get tested

annejan:
The releasing skill's FreeBSD step said "Test in a FreeBSD VM: `make stage check-plist stage-qa` and `portlint -AC`" and nothing about how. The how exists: [annejan/qtpass-freebsd-port-test](https://github.com/annejan/qtpass-freebsd-port-test), used for both the 1.8.0 and 1.8.1 port updates, and it has grown past what the skill knew — a `poudriere testport` job in clean jails (the one that catches a missing `*_DEPENDS`, which the pkg-preinstalled job cannot), and an OpenBSD `security/qtpass` job.

Skill now says: commit the diff as `patches/qtpass-X.Y.Z.patch` or dispatch the workflow with a ports branch; what it runs and on which releases/architectures; wait for green, then Bugzilla. Plus two facts a 2.0 port bump needs: the port lists the desktop file/metainfo/icons since 1.8.1, and 2.0 adds `share/man/man1/qtpass.1.gz`. And that OpenBSD is not ours — the tested diff goes to `ports@openbsd.org` with the maintainer in Cc.

Docs only. Prettier and super-linter clean.

## Summary by CodeRabbit

- **Documentation**
  - Expanded downstream release guidance to include OpenBSD packaging alongside FreeBSD.
  - Added platform-specific validation requirements, including FreeBSD build and dependency checks across supported architectures and versions.
  - Documented OpenBSD port checks for validation, building, packaging, and installation.
  - Clarified patch naming, branch, staged-file, committer-check, and submission requirements.
  - Updated guidance to require all platform checks to pass before submitting changes.


annejan:
Valid: the manual dispatch exposes `run_suite` and `false` skips the `make check` step the skill lists. Fixed in f23fb1f8c — the instructions now say to leave it at its default of `true`, and why.

The SkillSpector findings attached to the same comment are the generic scanner reacting to a release skill containing release commands (tags, uploads, a workflow dispatch); they describe what the document is for, not a defect in it, so no change for those.

PR #1769, opened 2026-09-16
parseDropData returns std::optional; one startProcess and one read path in Executor (#1682)

annejan:
Two section 4 items of #1682: **`parseDropData` → `std::optional`, name the drag MIME type once** (`src/storemodel.h:73`) and **De-duplicate `Executor::startProcess`/`startProcessBlocking` and the two read branches in `finished`** (`src/executor.cpp:32`, `:335`).

Pure refactor, no behaviour change: **+79 / −89**, 27 suites / 989 passed / 0 failed.

## `StoreModel`

- `parseDropData(const QMimeData *, dragAndDropInfoPasswordStore *out) -> bool` becomes `static parseDropData(const QMimeData *) -> std::optional<dragAndDropInfoPasswordStore>`, and does the null/`hasFormat` check itself. `canDropMimeData` had its own inline copy of the decode block instead of calling it; both it and `dropMimeData` now go through the one function.
- `"application/vnd.qtpass.dragAndDropInfoPasswordStore"` appeared 5× in `src/` and 6× in `tst_storemodel.cpp`. It is `kStoreDragMimeType` (an `inline constexpr QLatin1StringView`, no allocation, no ODR issue) in `storemodel.h`; every use, tests included, names it. Two `QVERIFY2` messages that string-concatenated the literal got plain text.

## `Executor`

- `startProcess(app, args)` (on `m_process`) and `startProcessBlocking(QProcess &, app, args)` were the same body with the same `"wsl "`-prefix handling — the spot that would silently drift the next time the WSL invocation changes (it already differs from `Pass::resolveGpgconfCommand`, which parses the prefix its own way). One static `startProcess(QProcess &, app, args)` serves the queued and the blocking path.
- `onProcessFinished` read stdout/stderr with the identical `readStdout` / `readStderr || exitCode != 0` block in both the `NormalExit` and the crash branch. `collectOutput(item, exitCode)` does it once; the `QT_DEBUG` trace of a non-zero exit stays where it was.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command execution error reporting by consistently capturing relevant standard error output for unsuccessful processes.
  * Ensured non-zero exit codes are logged reliably during normal process termination.

* **Refactor**
  * Unified queued and blocking command execution paths for more consistent behavior.
  * Improved drag-and-drop data validation and handling without changing supported functionality.

* **Tests**
  * Updated drag-and-drop coverage to use the shared MIME type definition.


annejan:
Valid as a consistency fix, applied in 471e63609: `main/main.pro` said `CONFIG += c++11` while `src/src.pro` and `tests/tests.pri` say `c++17`, and `main.cpp` reaches `storemodel.h` through `mainwindow.h`.

For the record it was not actually failing: Qt 6's qmake enforces a C++17 floor whatever a project requests (the generated `main/Makefile` carries no `-std=` at all, GCC 16's default is above 17), which is why every CI job was green. But a `.pro` that asks for a standard the code cannot be built with is a trap for the next reader, so it now says what is meant.

Issue #1842, opened 2026-09-19
Security review follow-ups: .gpg-id integrity, re-encryption, logging


annejan:
Findings from an external read of the 2.0 code (2026-09-19), each verified against `main` before it landed here. Checked items have a PR.

## Real, fixed or in progress

All below shipped in #1843 (`.gpg-id` integrity), #1844 (round two), #1845 (rollback/recovery), #1847 (`--` before paths), #1848 (symlinks are not entries); strings in #1846. Later rounds: #1849 (actions pinned, large-stdin test), #1850 (one WSL command parser), #1851 (junctions, linked folders, `removeTree`), #1852 (key generation: no passphrase only on request), #1853 (`.gpg-id` rewrite test), #1854 (executor tests on Windows CI), #1856 (Flatpak image digest, docs deploy job, Doxygen checksum), #1857 (atomic `ProfileInit` `.gpg-id`, path-boundary `getGpgIdPath`); strings in #1858.

- [x] **Verify-then-reread race on `.gpg-id`** — `Insert()` and `verifyGpgIdForDir()` verified the signature by path and then read the file again; whoever can write to the store in between picks the recipients. Fix: read once, `gpg --verify <sig> -` on those bytes, parse the same bytes; the re-encryption cache maps `.gpg-id` path → verified recipient list.
- [x] **`.gpg-id` and `.gpg-id.sig` committed separately** — an intermediate (or, when the second commit failed, permanent) history state with a new list under the old signature. Fix: one `git add`/`git commit` for both; no commit when nothing changed.
- [x] **`Init()` re-encrypted after the `.gpg-id` commit failed** — the working tree followed a recipient list the repository did not have. Fix: abort, report `processErrorExit`.
- [x] **Auto-pull failure treated as "continue"** even when the pull stopped in a merge. Fix: abort re-encryption when `git ls-files --unmerged` is non-empty; a pull that merely could not reach the remote still continues, as before.
- [x] **Signing key accepted anything** — `pass` matches `PASSWORD_STORE_SIGNING_KEY` against VALIDSIG fingerprints, so a short key ID signs and never verifies. Fix: the profile form requires 40/64-hex fingerprints.
- [x] **`emitProcessFinishedSignal` comment says deny-by-default, code is allow-by-default** — a new `PASS_*` value would reach `finishedAnyWithPid` (and the process output panel) unless someone remembers the deny list. Fix: allow-list + test.
[...]

PR #1864, opened 2026-09-20
A link inside the store is not part of it, at every operation

annejan:
## Summary

#1848/#1851 taught the *walks* (re-encryption, search, staging) to skip symlinks and junctions. The ordinary operations still followed them: the tree shows a linked entry like any other, and gpg, pass, QFile and QSaveFile all read and write through a link. A shared store's co-writer can make `git pull` create `Bank.gpg -> /elsewhere/x.gpg`; clicking it showed it, Edit would have re-encrypted whatever it pointed to with the store's recipients and committed it, a new entry saved into a linked folder landed outside the store, Copy read the target's bytes in.

## Rule

`Pass::refuseLinkedPath(path, includeSelf)` — an entry or folder that is, or lies behind, a link inside the store is refused with a message before anything runs, in **both backends**, for Show, Insert, Move, Copy (src, dest **and the resolved destination**), Init (the folder **and** its `.gpg-id`/`.gpg-id.sig` leaves — QSaveFile, `pass init` and `gpg --output` all write through a planted link). Remove refuses what lies behind a link; the link itself is unlinked by QtPass (`Util::removeTree`) and git is asked to forget it only if it tracked it — `pass rm -rf <folder>` turns the folder into `<folder>/` and rm then **empties the link's target** (reproduced), `git rm -rf -- link/` fails on the pathspec. The configured store root may still be a link (a synced folder is a normal setup).

Metadata: `getGpgIdPath` does not take a linked `.gpg-id` as a folder's list (the parent's applies), a linked root list reads as missing, and `GpgIdSigner` refuses a linked `.gpg-id` or `.gpg-id.sig` before `gpg --verify` — a validly signed pair cannot be replayed through two links.

UI: a refusal emits `startingExecuteWrapper`, `critical` and `processErrorExit`, i.e. exactly what an operation that starts and fails does, so the panel and browser clear, the OTP/edit paths release the interface (they used to sit disabled for the 30 s watchdog / at "Decrypting…"), and `onOtp()`'s fast path cannot copy the previous entry's code for the refused one.

## Process

Three commits; the second and third are what two adversarial review rounds (three lenses each, findings reproduced against the built library before being accepted) found in the first: `pass rm` following the link, Init's leaf, the dangling-link destination of a drop-copy, the armed UI, `git commit` on an untracked link, the browser keeping the previous entry, Init's un-normalised folder path from the context menu. The false-refusals lens built symlinked-root stores and found nothing refused wrongly.

## Tests

imitatepass +3 (every operation on a link/behind a link refused with gpg never run; Remove unlinks a link and leaves the target; linked `.gpg-id` semantics; git-tracked vs untracked link removal), realpass +1 (pass never sees a linked folder; ls-files decides), gpgidsigner +1 (linked pair never reaches gpg), mainwindow +1 (refused click clears the previous entry, releases the UI, OTP copies nothing), util (`isUnderLink(includeSelf=false)`). All suites green.

SECURITY.md gains the paragraph on shared stores and links. Refs #1842.

## Summary by CodeRabbit

* **Security Enhancements**
  * Blocked password-store operations involving symbolic links and NTFS junctions.
  * Prevented linked entries, folders, and metadata files from being followed or used for recipient and signature verification.
  * Linked folders are safely removed without affecting their targets; the configured store root may still be linked.

* **Bug Fixes**
  * Failed operations now clear pending display and copy state, release the interface, and prevent stale OTP values from being reused.
  * Search now scans real password-store files without following linked paths.

* **Documentation**
  * Added security guidance covering linked paths and shared password-store risks.


annejan:
Third commit (8dc8516) on top of the review:

- **`RealPass::Grep()`** — the external review's point holds: `pass grep` enumerates with `find -L`, which follows a link out of the store and decrypts what it finds. The pass backend's Search now uses the same `NativeGrep` as ImitatePass (real files only, via `Util::regularFilesUnder`), so both backends match alike; with no GPG executable configured it answers empty with a status message. Test: a real `folder/real.gpg` plus `Bank.gpg -> outside/secret.gpg`, a gpg stand-in that logs what it is handed — only the real entry is decrypted and found, pass itself is never invoked.
- The five CI failures were one new test asserting an exact `processErrorExit` count: on every runner but my machine the window's start-up gpg calls (real gpg, key listing) finish — and fail — inside the test's nested loops. It counts `critical()` only now and asserts the behaviour (panel cleared, UI enabled, OTP copies nothing).
- Lint: SECURITY.md paragraph split under the 400-column limit.

